PR #11, opened 2023-08-14
test

myarmolinsky:
Change-type: patch

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -245,7 +245,7 @@ export const theme = createTheme({
 			tertiary: '#8F9297',
 		},
 	},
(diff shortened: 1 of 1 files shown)


src/theme.ts
```diff
@@ -245,7 +245,7 @@ export const theme = createTheme({
 			tertiary: '#8F9297',
 		},
 	},
-	spacing: [0, 4, 8, 16, 32, 64, 128],
+	spacing: [0, 4, 8, 16, 32, 64, 129],
```
JSReds, line 248: Absolutelly not a good change XP

PR #13, opened 2023-08-22
Add DialogWithCloseButton component and Dialog styling

myarmolinsky:
Add DialogWithCloseButton component and Dialog styling

Connects-to: https://github.com/balena-io/balena-ui/pull/6072
Change-type: minor

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -256,6 +256,29 @@ export const theme = createTheme({
 				},
 			},
 		},
+		MuiDialogTitle: {
+			styleOverrides: {
+				root: {
+					padding: '24px',
+					fontSize: '24px',
+				},
+			},
+		},
+		MuiDialogContent: {
+			styleOverrides: {
+				root: {
+					padding: '0px 24px',
+					margin: '0px 0px 24px',
+				},
+			},
+		},
+		MuiDialogActions: {
+			styleOverrides: {
+				root: {
+					padding: '0px 24px 24px',
+				},
+			},
+		},
 		MuiCard: {
 			styleOverrides: {
 				root: {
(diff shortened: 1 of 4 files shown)


src/Material/DialogWithCloseButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+					{title}
+					{onClose ? (
+						<IconButton
+							aria-label="close"
+							onClick={(e) => {
+								onClose(e, 'backdropClick');
+							}}
+							sx={{
+								color: (theme) => theme.palette.grey[500],
+								ml: 3,
```
JSReds, line 35: just to make sure, is this necessary ? I thought the space between rule would be enough ( I might be totally wrong

src/theme.ts
```diff
@@ -256,6 +256,32 @@ export const theme = createTheme({
 				},
 			},
 		},
+		MuiDialogTitle: {
+			styleOverrides: {
+				root: {
+					padding: '24px',
+					fontSize: '24px',
+					fontFamily: '"Source Sans Pro", Helvetica, sans-serif',
```
JSReds, line 264: This should not be necessary, as all strings should live inside a Typography component. Having said that if really needed it should be possible to make a generic rule, this might be of help https://mui.com/material-ui/react-css-baseline/

drskullster: @JSReds I agree that the font-family rule should be generic but we can keep it like this for now, I'll add the css baseline (probably scoped as not to interfere too much with Grommet) in a separate PR.

src/Material/DialogWithCloseButton.tsx
myarmolinsky: Without it, a long enough title could touch the X: 
image

drskullster:
@myarmolinsky there's quite a big gap between the dialog heading and the title. I don't see it in the screenshots on the UI PR, did you add the spacing recently?

image

myarmolinsky:
LGTM

> @myarmolinsky there's quite a big gap between the dialog heading and the title. I don't see it in the screenshots on the UI PR, did you add the spacing recently?
> 
> image

Yeah the spacing changed after those screenshots. It's an iffy spot because the gap increases because DialogTitle and DialogContent are stacked. Let me think about a solution for this 🤔

myarmolinsky:
LGTM

PR #17, opened 2023-09-06
Add DownloadImageModal component

JSReds:
Change-type: minor

Screenshot 2023-11-29 at 17 28 33

Screenshot 2023-11-29 at 17 28 50

Screenshot 2023-11-29 at 17 29 19

Screenshot 2023-11-29 at 17 28 22


JSReds:
sorry for removing the contribution @drskullster 🙏 🙏 🙏 🙏

src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -0,0 +1,211 @@
...
+	}) => {
+		const [currentOs, setCurrentOs] = useState<OsOptions>(getUserOs());
+
+		const instructions = useMemo(() => {
+			if (
+				deviceType?.instructions == null ||
+				Array.isArray(deviceType.instructions) ||
+				typeof deviceType.instructions !== 'object'
+			) {
+				return deviceType?.instructions;
```
myarmolinsky, line 54:
```suggestion
				return deviceType.instructions;
```

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -0,0 +1,503 @@
...
+			const selectedValueIsPreferred = preferredSelectionOpts.some(
+				(ver) => ver.value === version.value,
+			);
+			if (selectedValueIsPreferred) {
+				return;
+			}
+
+			const preferred =
+				preferredSelectionOpts.find((ver) => ver.isRecommended) ??
+				preferredSelectionOpts?.[0];
```
myarmolinsky, line 142: Optional chaining in one but not the other? Do we need the optional chain or not? If yes I would expect it on both, if no I would expect it on neither

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -0,0 +1,503 @@
...
+
+		const handleSelectedDeviceTypeChange = useCallback(
+			(deviceType: DeviceType) => {
+				if (model.deviceType.slug === deviceType.slug) {
+					return;
+				}
+
+				const newDeviceType = compatibleDeviceTypes?.find(
+					(cdt) => cdt.slug === deviceType.slug,
+				);
```
myarmolinsky, line 156:
```suggestion
				if (model.deviceType.slug === deviceType.slug || !compatibleDeviceTypes) {
					return;
				}

				const newDeviceType = compatibleDeviceTypes.find(
					(cdt) => cdt.slug === deviceType.slug,
				);
```

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -0,0 +1,503 @@
...
+				autoComplete="off"
+				p={2}
+				ref={formElement}
+			>
+				<input type="hidden" name="_token" value={authToken} />
+				<input type="hidden" name="appId" value={applicationId} />
+				<input type="hidden" name="fileType" value=".zip" />
+				<Grid py={3} display="flex" flexWrap="wrap" gap={2}>
+					{compatibleDeviceTypes && compatibleDeviceTypes.length > 1 && (
+						<Grid display="flex" flexDirection="column" flex="1">
```
myarmolinsky, line 208: Why are we using `Grid` here instead of `Box`? I don't see us using any of the alignment properties that Grid grants us

Also I see that this is Gridv1, not Gridv2 like as we agreed to stick to

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -0,0 +1,503 @@
...
+);
+
+const DeviceTypeItem: React.FC<{ deviceType: DeviceType }> = ({
+	deviceType,
+}) => {
+	return (
+		<MenuItem value={deviceType.slug}>
+			<Avatar
+				variant="square"
+				src={deviceType.logo ?? undefined}
```
myarmolinsky, line 441: Shouldn't we have the default logo if the `deviceType.logo` is nullish?

src/components/DownloadImageDialog/index.tsx
```diff
@@ -0,0 +1,523 @@
...
+			formModel.version,
+			setDownloadSize,
+		);
+	}, [formModel.deviceType, getDownloadSize, formModel.version]);
+
+	useEffect(() => {
+		if (!compatibleDeviceTypes || !getSupportedOsVersions) {
+			return;
+		}
+		// const applicationType = getExpanded(application.application_type);
```
myarmolinsky, line 355: ?

src/components/DownloadImageDialog/index.tsx
```diff
@@ -0,0 +1,523 @@
...
+		[setFromModel],
+	);
+
+	return (
+		<Dialog open={open} onClose={onClose} maxWidth="lg" fullWidth sx={{ p: 4 }}>
+			<DialogTitle
+				display="flex"
+				justifyContent="space-between"
+				alignItems="center"
+				variant="h5"
```
myarmolinsky, line 434: Do we need this variant here if the Typography has it?

src/components/DownloadImageDialog/utils.ts
```diff
@@ -0,0 +1,37 @@
+import template from 'lodash/template';
+import { Dictionary, OptionalNavigationResource, OsTypesEnum } from './models';
+
+export const FALLBACK_LOGO_UNKNOWN_DEVICE =
+	'https://dashboard.balena-cloud.com/img/unknown-device.svg';
```
myarmolinsky, line 5: This really relies on us not changing the name of that file 😕 
Perhaps we should just add the SVG to ui-shared-components as well?

src/components/DownloadImageDialog/version.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const LEGACY_OS_VERSION_MAJOR = 1;
+
+// This returns the 3 most preferred versions for an os type. For multi-line os types, that would be the latest of each line, otherwise it is the latest 3 versions.
+export const getPreferredVersionOpts = (
+	versionOpts: VersionSelectionOptions[],
+) => {
+	const [supportedVersions, legacyVersions] = partition(versionOpts, (v) => {
+		// TODO: check if worth installing semver on rendition;
+		// const major = semver.major(v.strippedVersion);
```
myarmolinsky, line 73: 🚓

src/components/DropDownButton/index.tsx
```diff
@@ -89,6 +89,7 @@ export const DropDownButton: React.FC<DropDownButtonProps> = ({
 					aria-label="actions"
 					aria-haspopup="menu"
 					onClick={handleToggle}
+					sx={(theme) => ({ pl: 2, pr: `calc(${theme.spacing(2)} + 2px)` })}
```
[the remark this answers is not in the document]
myarmolinsky: Sure, let's add a comment

src/components/DownloadImageDialog/ApplicationInstructions.tsx
JSReds: we need to keep it as if the first rule in the if is true then it would explod : 

if (
		deviceType?.instructions == null ||
		Array.isArray(deviceType.instructions) ||
		typeof deviceType.instructions !== 'object'
	)
	
	deviceType?.instructions == null might be that deviceType is null. We just do that to be sure that we return the exact same type

src/components/DownloadImageDialog/utils.ts
JSReds: I'm still not sure about this point, I'll add a TODO to verify how we can achieve it without duplicating the image

.storybook/preview.tsx
```diff
@@ -32,7 +32,9 @@ const preview: Preview = {
 	decorators: [
 		(Story) => (
 			<ThemeProvider theme={theme}>
-				<Story />
+				<ScopedCssBaseline>
```
myarmolinsky, line 35: Just to make sure, I believe @drskullster mentioned he was going to add this in a standalone PR. Was using this agreed upon?

src/components/ButtonWithTracking/index.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
-import { Button, ButtonProps } from '@mui/material';
+import { Button, ButtonProps, Tooltip } from '@mui/material';
 import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
 
 export interface ButtonWithTrackingProps extends ButtonProps {
 	eventName: string;
 	eventProperties?: { [key: string]: any };
+	tooltip?: string;
```
myarmolinsky, line 8: Not a strong feeling about this, but perhaps we should not use a tooltip prop and instead just wrap any Button we want to have a tooltip with Tooltip manually? Just concerned of whether this is the beginning of creating complex custom components as in rendition (I can be overreacting lol)

src/components/DownloadImageDialog/ApplicationInstructions.tsx
myarmolinsky: You're right, my bad

src/components/OrderedListItem/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const OrderedListItem: React.FC<OrderedListItemProps> = ({
+	index,
+	children,
+	...orderedListItemProps
+}) => {
+	return (
+		<ListItem
+			sx={{ pl: 0, ...orderedListItemProps.sx }}
+			{...orderedListItemProps}
```
myarmolinsky, line 22: Doesn't this make it so that the sx is overriden by the subsequent `{...orderedListItemProps}` if it has its own `sx` since we did not destructure the `sx` from it?
```suggestion
			sx={{ pl: 0, ...orderedListItemProps.sx }}
			{...omit(orderedListItemProps, 'sx')}
```
(preferably just destructure the sx above, I just used the `omit` to communicate what I am taking about :P)

src/components/DownloadImageDialog/version.ts
```diff
@@ -0,0 +1,113 @@
...
+			hasPrebuiltVariants: true;
+			rawVersions: {
+				dev?: string;
+				prod?: string;
+			};
+	  }
+);
+
+export const transformVersions = (versions: OsVersion[]) => {
+	// Get a single object per stripped version, with both variants of it included (if they exist). It expects a sorted `
```
myarmolinsky, line 27: > It expects a sorted `

Perhaps I am just not understanding, is this a complete thought? 😓

src/components/DownloadImageDialog/version.ts
```diff
@@ -0,0 +1,113 @@
...
+
+		optsByVersion[version.strippedVersion] = {
+			title,
+			value: version.strippedVersion,
+			osType: version.osType,
+			line: version.line,
+			isRecommended: version.isRecommended,
+			knownIssueList: version.known_issue_list,
+			// Unified releases in the model have variant === ''
+			// but we also test for nullish for backgwards compatibility w/ the typings.
```
myarmolinsky, line 44: Nit 😅 
```suggestion
			// but we also test for nullish for backwards compatibility w/ the typings.
```

src/components/DownloadImageDialog/version.ts
```diff
@@ -0,0 +1,113 @@
...
+};
+
+const LEGACY_OS_VERSION_MAJOR = 1;
+
+// This returns the 3 most preferred versions for an os type. For multi-line os types, that would be the latest of each line, otherwise it is the latest 3 versions.
+export const getPreferredVersionOpts = (
+	versionOpts: VersionSelectionOptions[],
+) => {
+	const [supportedVersions, legacyVersions] = partition(versionOpts, (v) => {
+		// TODO: check if worth installing semver on ui-shared-components;
```
myarmolinsky, line 72: Curious, why might it not be worth it?

src/components/DownloadImageDialog/version.ts
```diff
@@ -0,0 +1,113 @@
...
+			if (
+				match &&
+				!preferredDefaultOpts.find((v) => v.value.startsWith(match[0]))
+			) {
+				preferredDefaultOpts.push(option);
+			}
+		}
+
+		return preferredDefaultOpts;
+	}
```
myarmolinsky, line 112: The `if` statement above returns, no need to wrap this code in an `else`
```suggestion
	}
		const preferredDefaultOpts: VersionSelectionOptions[] = [];
		for (const option of opts) {
			if (preferredDefaultOpts.length >= 3) {
				break;
			}

			const match = option.value.match(/\d+\.\d+\./);
			if (
				match &&
				!preferredDefaultOpts.find((v) => v.value.startsWith(match[0]))
			) {
				preferredDefaultOpts.push(option);
			}
		}

		return preferredDefaultOpts;
```

src/components/DownloadImageDialog/models.ts
```diff
@@ -0,0 +1,139 @@
+export interface Dictionary<T> {
```
myarmolinsky, line 1: Just curious, I don't want to complicate this PR further so we can leave it as is regardless of your answer:

Did we just bring these typings over from rendition or did you go through them all and filter out anything we no longer need?

src/components/DownloadImageDialog/VariantSelector.tsx
```diff
@@ -0,0 +1,122 @@
...
+							title={
+								isDisabled
+									? 'This edition is not available for the selected version'
+									: undefined
+							}
+							key={index}
+						>
+							<Box display="flex" flexDirection="column">
+								<FormControlLabel
+									sx={{ opacity: isDisabled ? 0.6 : 1 }}
```
myarmolinsky, line 60: Is 0.6 the opacity we use everywhere else? Asking because I believe I've seen us using 0.4 in other places (but it could easily have been for a different case)

src/components/DownloadImageDialog/VariantSelector.tsx
```diff
@@ -0,0 +1,122 @@
...
+													<Lightbulb sx={{ width: 14, height: 14 }} />
+													Recommended for first time users
+												</Typography>
+											</Box>
+										) : (
+											label
+										)
+									}
+								/>
+								{isDev ? (
```
myarmolinsky, line 88: We should handle this with a dictionary by passing a "title" with "description" instead of needing a ternary to manually check which description to show depending on `isDev`

src/components/ButtonWithTracking/index.tsx
JSReds: That's what I was initially doing, but i discovered that this can't be done without having to manipulate all our components. We should make all components wrapped in a forwardRef, which I honestly don't really like. So to make it work, we need to use Tooltip to only wrap MUI components.

src/components/DownloadImageDialog/version.ts
JSReds: nice catch, that was an old comment, not sure where it come from, btw it was probably a sorted `OsVersion array`. I'll remove it as it is not really necessary.

src/components/DownloadImageDialog/version.ts
JSReds: not sure it is valuable to install an entire library to just remove a single line of code. I would first check if we can also use it to improve other pieces.

src/components/DownloadImageDialog/VariantSelector.tsx
JSReds: we should make a generic "disabled" rule in the theme adding it in design-tokens first. For now I've switched to 0.4 which is probably better.

src/components/DownloadImageDialog/version.ts
myarmolinsky: I was thinking it might count as a shared component since we use it in the UI, but I see now we don't use it in hub so it doesn't count as shared yet

src/components/DownloadImageDialog/models.ts
JSReds: I should have removed all unnecessary types. Next step would be to unify what is used in multiple places

src/components/DownloadImageDialog/VariantSelector.tsx
myarmolinsky: Agree

src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -0,0 +1,231 @@
...
+				if (!oses.includes(currentOs) && oses.length > 0) {
+					setCurrentOs(oses[0] as OsOptions);
+				}
+			}
+		}, [currentOs, setCurrentOs, instructions, hasOsSpecificInstructions]);
+
+		if (!deviceType || !instructions) {
+			return (
+				<Typography variant="body1">
+					Instructions for this application are not currently available. Please
```
myarmolinsky, line 91: Should it be `application` or `device type`?
```suggestion
					Instructions for this device type are not currently available. Please
```

myarmolinsky (approved):

PR #23, opened 2023-09-29
Add form component

JSReds:
Change-type: minor

### DIFF
--- a/src/components/Form/Form.stories.tsx
+++ b/src/components/Form/Form.stories.tsx
@@ -1,0 +1,103 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { RJSForm } from '.';
+
+const meta = {
+	title: 'Other/RJSForm',
+	component: RJSForm,
+	tags: ['autodocs'],
+} satisfies Meta<typeof RJSForm>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+	args: {
+		schema: {
+			title: 'A registration form',
+			description: 'A simple form example.',
+			type: 'object',
+			required: ['firstName', 'lastName'],
+			properties: {
+				firstName: {
+					type: 'string',
+					title: 'First name',
+					default: 'Chuck',
+				},
+				lastName: {
+					type: 'string',
+					title: 'Last name',
+				},
+				age: {
+					type: 'integer',
+					title: 'Age',
+				},
+				bio: {
+					type: 'string',
+					title: 'Bio',
+				},
+				password: {
+					type: 'string',
+					title: 'Password',
+					minLength: 3,
+				},
+				passwordStrength: {
+					type: 'string',
+					title: 'Password make it strong',
+					minLength: 3,
+				},
+				telephone: {
+					type: 'string',
+					title: 'Telephone',
+					minLength: 10,
+				},
+			},
+		},
+		uiSchema: {
+			firstName: {
+				'ui:autofocus': true,
+				'ui:emptyValue': '',
+				'ui:placeholder':
+					'ui:emptyValue causes this field to always be valid despite being required',
+				'ui:autocomplete': 'family-name',
+				'ui:enableMarkdownInDescription': true,
+				'ui:description':
+					'Make text **bold** or *italic*. Take a look at other options [here](https://probablyup.com/markdown-to-jsx/).',
+			},
+			lastName: {
+				'ui:autocomplete': 'given-name',
+				'ui:enableMarkdownInDescription': true,
+				'ui:description':
+					'Make things **bold** or *italic*. Embed snippets of `code`. <small>And this is a small texts.</small> ',
+			},
+			age: {
+				'ui:widget': 'updown',
+				'ui:title': 'Age of person',
+				'ui:description': '(earth year)',
+			},
+			bio: {
+				'ui:widget': 'textarea',
+			},
+			password: {
+				'ui:widget': 'password',
+			},
+			passwordStrength: {
+				'ui:widget': 'password',
+				'ui:help': 'Hint: Make it strong!',
+				'ui:options': {
+					showPasswordStrengthMeter: true,
+				},
+			},
+			telephone: {
+				'ui:options': {
+					inputType: 'tel',
+				},
+			},
+		},
+		submitButtonProps: {
+			sx: {
+				display: 'flex',
+				alignSelf: 'flex-end',
+			},
+		},
+	},
+};
--- a/src/components/Form/Widgets/PasswordWidget.tsx
+++ b/src/components/Form/Widgets/PasswordWidget.tsx
@@ -1,0 +1,126 @@
+import {
+	Box,
+	FormControl,
+	IconButton,
+	InputAdornment,
+	TextField,
+	Typography,
+} from '@mui/material';
+import * as React from 'react';
+import { Visibility, VisibilityOff } from '@mui/icons-material';
+import { WidgetProps } from '@rjsf/utils';
+import zxcvbn from 'zxcvbn';
+
+const STRENGTH_TITLES = ['Very weak', 'Weak', 'Good', 'Strong', 'Very strong'];
+const STRENGTH_STYLES = [
+	{ width: 0 },
+	{ width: '25%', backgroundColor: 'orange' },
+	{ width: '50%', backgroundColor: 'yellow' },
+	{ width: '75%', backgroundColor: 'green' },
+	{ width: '100%', backgroundColor: 'darkgreen' },
+];
+
+export interface PasswordStrengthProps {
+	password?: string;
+}
+
+const PasswordStrength = ({ password }: PasswordStrengthProps) => {
+	const [strengthScore, setStrengthScore] = React.useState<
+		number | undefined
+	>();
+
+	React.useEffect(() => {
+		if (password) {
+			try {
+				const { score } = zxcvbn(password);
+				setStrengthScore(score);
+			} catch {
+				// Ignore any errors, as we only want to show the strength meter if it is available.
+			}
+		}
+	}, [password]);
+
+	if (!password || strengthScore === undefined) {
+		return null;
+	}
+
+	return (
+		<Box>
+			<Typography>
+				Password strength: <em>{STRENGTH_TITLES[strengthScore]}</em>
+			</Typography>
+			<Box
+				sx={[
+					STRENGTH_STYLES[strengthScore],
+					{ height: '4px', transition: 'all ease-out 150ms' },
+				]}
+			/>
+		</Box>
+	);
+};
+
+export const PasswordWidget = ({
+	id,
+	label,
+	value,
+	disabled,
+	name,
+	placeholder,
+	onBlur,
+	onChange,
+	onFocus,
+	options,
+	required,
+}: WidgetProps) => {
+	const [showPassword, setShowPassword] = React.useState(false);
+
+	const change = ({ target: { value } }: any) => {
+		return onChange(value === '' ? options.emptyValue : value);
+	};
+
+	return (
+		<FormControl>
+			<TextField
+				label={label}
+				required={required}
+				id={id}
+				type={showPassword ? 'text' : 'password'}
+				{...{
+					value,
+					disabled,
+					name,
+					placeholder,
+				}}
+				onChange={change}
+				onBlur={
+					onBlur &&
+					((event: React.FocusEvent<HTMLInputElement, Element>) =>
+						onBlur(id, event.target.value))
+				}
+				onFocus={
+					onFocus &&
+					((event: React.FocusEvent<HTMLInputElement, Element>) =>
+						onFocus(id, event.target.value))
+				}
+				InputProps={{
+					endAdornment: (
+						<InputAdornment position="end">
+							<IconButton
+								onClick={() => setShowPassword((show) => !show)}
+								onMouseDown={(event: React.MouseEvent<HTMLButtonElement>) => {
+									event.preventDefault();
+								}}
+								edge="end"
+							>
+								{showPassword ? <VisibilityOff /> : <Visibility />}
+							</IconButton>
+						</InputAdornment>
+					),
+				}}
+			/>
+			{options.showPasswordStrengthMeter && (
+				<PasswordStrength password={value} />
+			)}
+		</FormControl>
+	);
+};
--- a/src/components/Form/index.tsx
+++ b/src/components/Form/index.tsx
@@ -1,0 +1,76 @@
+import Form from '@rjsf/mui';
+import ajvValidator from '@rjsf/validator-ajv8';
+import { FormProps as RjsFormProps } from '@rjsf/core';
+import { Box, BoxProps, Button, ButtonProps } from '@mui/material';
+import { PasswordWidget } from './Widgets/PasswordWidget';
+import { SelectWidget } from './Widgets/SelectWidget';
+import { Fragment } from 'react';
+import { ObjectFieldTemplate } from './FieldTemplates/ObjectFieldTemplate';
+
+const internalWidgets: {
+	[k: string]: any;
+} = {
+	PasswordWidget,
+	SelectWidget,
+};
+
+export interface RJSFormProps
+	extends Omit<RjsFormProps, 'validator' | 'onFocus' | 'onBlur'>,
+		Pick<BoxProps, 'sx' | 'onFocus' | 'onBlur'>,
+		Partial<Pick<RjsFormProps, 'validator'>> {
+	/** If true, do not display the form submit button */
+	hideSubmitButton?: boolean;
+	/** Properties that are passed to the submit button, these are the same props used for the [`Button`](#button) component */
+	submitButtonProps?: ButtonProps;
+	/** If passed, it will show a secondary button, these are the same props used for the [`Button`](#button) component */
+	actionButtons?: ButtonProps[];
+}
+
+export const RJSForm: React.FC<React.PropsWithChildren<RJSFormProps>> = ({
+	hideSubmitButton,
+	submitButtonProps,
+	actionButtons,
+	validator = ajvValidator,
+	widgets,
+	fields,
+	children,
+	ref,
+	sx,
+	onFocus,
+	onBlur,
+	templates,
+	...otherProps
+}) => {
+	return (
+		<Box sx={sx} onFocus={onFocus} onBlur={onBlur}>
+			<Form
+				ref={ref}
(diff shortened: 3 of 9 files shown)


JSReds (commented):
I know that TextWidget, TextareaWidget, UpDownWidget and DateWidget might be merged in a single widget, but I really don't like the amount of if statements and the way the if looks like. I found this way much more readable. Since RJSF choose the widget based on the schema.type (https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.1) I would keep their logic to choose the right widget.

src/components/Form/DateTimeWidget.tsx
```diff
@@ -0,0 +1,49 @@
...
+}: WidgetProps) => {
+	const change = ({ target: { value } }: any) => {
+		return onChange(value === '' ? options.emptyValue : value);
+	};
+	return (
+		<FormControl variant="outlined">
+			<label htmlFor={id}>
+				{label}
+				{required ? '*' : null}
+			</label>
```
myarmolinsky, line 29: Can we please use this: https://mui.com/material-ui/api/input-label/#InputLabel-prop-required

src/components/Form/DateTimeWidget.tsx
```diff
@@ -0,0 +1,49 @@
...
+				{...{
+					value,
+					disabled,
+					name,
+					readonly,
+					autoFocus,
+					placeholder,
+					hideLabel,
+					multiple,
+				}}
```
myarmolinsky, line 42: Why did we specify the full object to destructure instead of just using ...props? For explicitness?

src/components/Form/DateTimeWidget.tsx
```diff
@@ -0,0 +1,49 @@
...
+				onChange={change}
+				onBlur={onBlur && ((event: any) => onBlur(id, event.target.value))}
+				onFocus={onFocus && ((event: any) => onFocus(id, event.target.value))}
```
myarmolinsky, line 45: `any`?

src/components/Form/PasswordWidget.tsx
```diff
@@ -0,0 +1,145 @@
...
+	password?: string;
+}
+
+const PasswordStrength = ({ password }: PasswordStrengthProps) => {
+	const [strengthScore, setStrengthScore] = React.useState<
+		number | undefined
+	>();
+
+	React.useEffect(() => {
+		if (zxcvbn && password) {
```
myarmolinsky, line 54: Why does `zxcvbn` need to be checked?

src/components/Form/PasswordWidget.tsx
```diff
@@ -0,0 +1,145 @@
...
+				setStrengthScore(score);
+			} catch {
+				// Ignore any errors, as we only want to show the strength meter if it is available.
+			}
+		}
+	}, [password]);
+
+	if (!password || strengthScore === undefined) {
+		return null;
+	}
```
myarmolinsky, line 66: Do we need both this and the `width: 0` in `STRENGTH_STYLES`?

src/components/Form/PasswordWidget.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+	return (
+		<PasswordStrengthContainer>
+			<PasswordStrengthText>
+				Password strength: <em>{STRENGTH_TITLES[strengthScore]}</em>
+			</PasswordStrengthText>
+			<PasswordStrengthMeter
+				style={STRENGTH_STYLES[strengthScore]}
+			></PasswordStrengthMeter>
```
myarmolinsky, line 75: ?
```suggestion
			<PasswordStrengthMeter
				style={STRENGTH_STYLES[strengthScore]}
			/>
```

src/components/Form/PasswordWidget.tsx
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+	}, [password]);
+
+	if (!password || strengthScore === undefined) {
+		return null;
+	}
+
+	return (
+		<PasswordStrengthContainer>
```
myarmolinsky, line 69: Perhaps we can use this and add `joy-ui` to our toolkit? https://mui.com/joy-ui/react-input/#password-meter

src/components/Form/DateTimeWidget.tsx
JSReds: Not sure why I left that version, thank you for pointing it out

src/components/Form/DateTimeWidget.tsx
JSReds: we want to specify them as doing {...props} will also add properties that we don't want to pass to the input as props are of type widgetProps which also include schema, uiSchema, registry, rawErrors...

src/components/Form/DateTimeWidget.tsx
JSReds: 🤦

src/components/Form/PasswordWidget.tsx
JSReds: yep as if strength score is 0 we still want to show "strength score: "

src/components/Form/PasswordWidget.tsx
JSReds: for sure I can improve the code btw :) I'll push a better version

src/components/Form/DateTimeWidget.tsx
myarmolinsky: > Our label will not make the MUI animation

It doesn't need to though? It just needs to be displayed 🤔 Can't we just disable the animation and still use InputLabel?

src/components/Form/PasswordWidget.tsx
myarmolinsky: Fair! And thanks for pushing the improved code :)

src/components/Form/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+	hideSubmitButton?: boolean;
+	/** Properties that are passed to the submit button, these are the same props used for the [`Button`](#button) component */
+	submitButtonProps?: ButtonProps;
+	/** If passed, it will show a secondary button, these are the same props used for the [`Button`](#button) component */
+	secondaryButtonProps?: ButtonProps;
+
+	validator?: RjsFormProps['validator'];
+}
+
+export const MuiForm: React.FC<React.PropsWithChildren<MuiFormProps>> = ({
```
myarmolinsky, line 36: Why name it `MuiForm` instead of just `Form`? I would prefer `RJSForm`, especially in the unlikely/faraway scenario that we switch from MUI

src/components/Form/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+				{hideSubmitButton && <span />}
+
+				{!hideSubmitButton && (
+					<Button
+						{...submitButtonProps}
+						color="customBlue"
+						variant="contained"
+						type="submit"
+					/>
+				)}
```
myarmolinsky, line 74: We may need to talk about this a bit, cc @drskullster 
I know this is what we already had in rendition but if you have it this way then the form's `onSubmit` does not fire with the `enter` button when the submit button is hidden (since it is technically not rendered). It's better to have `display: 'none'` when hidden IMO. However, I believe Jon noted that this may cause issues with accessibility. We should find a way forward that enables usage of native browser behavior (i.e. submit form on enter)

src/components/Form/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+				{!!secondaryButtonProps && (
+					<Button
+						{...secondaryButtonProps}
+						variant="outlined"
+						color="customBlue"
+					/>
+				)}
```
myarmolinsky, line 81: What's the case where we would have this?
I see that we use submitButtonProps and submitButtonText, but the only use I see of this is on the Profile setup page. I don't believe a form will generally have a second button (maybe there's some standard UX rules/requirements I'm missing here cc @drskullster, the only exception I can think of is a multi-page form where the secondary button might be to go back a page). Arguably the Profile setup page can become a `DialogWithCloseButton` or feature some other way to skip (i.e. just change the Submit button text to `Skip` if the inputs are unchanged maybe? 🤷 ) and we can avoid this extremely rare extra prop

myarmolinsky: >  I don't believe a form will generally have a second button (maybe there's some standard UX rules/requirements I'm missing here cc @drskullster, the only exception I can think of is a multi-page form where the secondary button might be to go back a page). 

Another example I can imagine is a `Reset to default` button, which would indeed be a valid usecase. But in terms of how we already use this prop, I'm not sure a `Skip` button is standard UX for a Form. Could be wrong though, I can see it go either way now tbh but I want us to think about it before just sticking with what we had last time

myarmolinsky, line 81: These two Buttons need to be reversed so that the primary (submit) button is on the right (which lines up with our Dialog UX)

src/components/Form/Form.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+const meta = {
+	title: 'Other/MuiForm',
+	component: MuiForm,
+	tags: ['autodocs'],
+} satisfies Meta<typeof MuiForm>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
```
myarmolinsky, line 13: Your submit button comes out as a small textless button (same with the secondary button if you try to add it): 
image

src/components/Form/index.tsx
myarmolinsky: Also I think the buttons should align right
Not sure what standard/ideal UX is on this point though

src/components/Form/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+	...otherProps
+}) => {
+	const handleKeyUpDown = useCallback(
+		(e: React.KeyboardEvent<HTMLDivElement>) => {
+			if (e.key === 'Enter' && !hideSubmitButton) {
+				e.stopPropagation();
+			}
+		},
+		[hideSubmitButton],
+	);
```
myarmolinsky, line 53: As noted in the comment about `hideSubmitButton`, we should discuss the appropriate way to enable submit on enter with native behavior instead of these workarounds we're making (I know this comes from rendition, just want to improve and do things right moving forward)

src/components/Form/DateTimeWidget.tsx
JSReds: the label style is completely different, we would need to create a styled component to make the style match what we need, it is not worth. The label without animation is still not what we want check image 
Screenshot 2023-11-02 at 12 48 34

src/components/Form/index.tsx
JSReds: I would not use Form as MUI might create a Form component at some point which would create confusion (no conflict but still confusing IMHO). I can go with RJSF but seems strange as it would be react-json-schema-form-form

src/components/Form/index.tsx
JSReds: @myarmolinsky I was keeping as a first step the logic we have in rendition to not make a complete change in a single PR. Ideally we should have a buttons.map here, not be opinionated. A form might have a clear button, a validate button, a submit button...  I'm trying to not focus on our single use case.

src/components/Form/index.tsx
JSReds: It was already in my TODO list to discuss this point today during the call. This is exactly why I've added the children property, as all form actions should be handled inside the Form.

src/components/Form/index.tsx
JSReds: This is not needed for the Form itself, form submit button would work with this without problems. I pretty sure this was added at some point as we have the dashboard showing a modal with a components that override this behaviour. We can try to remove it and try to fix the dashboard where necessary

src/components/Form/Form.stories.tsx
JSReds: thanks for noticing, just forgot to add a default value after removing the not needed property submitButtonText

src/components/Form/index.tsx
myarmolinsky: `RJSForm` :)

src/components/Form/index.tsx
myarmolinsky: > This is not needed for the Form itself, form submit button would work with this without problems. 

Doesn't work when `hideSubmitButton` is true. So this useEffect is used for that case as well. As noted earlier, this shouldn't be a case either ...

myarmolinsky: > We can try to remove it and try to fix the dashboard where necessary

We should 💯

src/components/Form/DateTimeWidget.tsx
drskullster: @JSReds I don't think that's too much work tbh, here's what I did in the theme to make it look ok:

```ts
		MuiInputLabel: {
			defaultProps: {
				shrink: true,
			},
			styleOverrides: {
				root: {
					transform: 'none',
					position: 'relative',
				},
			},
		},
```

The `sx` as well as the `shrink` props could be set in the theme so we wouldn't need to worry

drskullster: I can update that in a commit if you want?

src/components/Form/index.tsx
drskullster: @myarmolinsky @JSReds we can talk about it in our call but the TLDR; is that I'm ok with this implementation, because we can have external submit buttons very easily:

```html
<form id="my-form">
  <input type="text" />
</form>
<p>
  <button form="my-form" type="submit">External submit</button> <!-- works! -->
</p>
```

I'm wondering if we could make without the `<span/>` element though, any idea why this would be required?

myarmolinsky: But the user shouldn't have to use an external submit button, right? Shouldn't we just have the proper form support from the getgo?

src/components/Form/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+export interface RJSFormProps
+	extends Omit<RjsFormProps, 'validator' | 'onFocus' | 'onBlur'> {
+	/** If true, do not display the form submit button */
+	hideSubmitButton?: boolean;
+	/** Properties that are passed to the submit button, these are the same props used for the [`Button`](#button) component */
+	submitButtonProps?: ButtonProps;
+	/** If passed, it will show a secondary button, these are the same props used for the [`Button`](#button) component */
+	actionButtons?: ButtonProps[];
+
+	validator?: RjsFormProps['validator'];
```
myarmolinsky, line 25: I don't see us using this property in the UI, can we remove this?

src/components/Form/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+	/** allow styling wrapper container */
+	sx?: BoxProps['sx'];
+	onFocus?: BoxProps['onFocus'];
+	onBlur?: BoxProps['onBlur'];
```
myarmolinsky, line 29:
```suggestion
	extends Omit<RjsFormProps, 'validator' | 'onFocus' | 'onBlur'> & Pick<BoxProps, 'sx' | 'onFocus' | 'onBlur'>
```

src/components/Form/index.tsx
myarmolinsky: Or is the point solely to make this prop optional? If so then maybe the following as a nit? Up to you:
```suggestion
	extends Omit<RjsFormProps, 'validator' | 'onFocus' | 'onBlur'> & Partial<Pick<RjsFormProps, 'validator'>>
```

src/components/Form/SelectWidget.tsx
```diff
@@ -0,0 +1,71 @@
...
+				sanitizedEnumDisabled != null &&
+				sanitizedEnumDisabled.includes(option.value),
+		};
+	});
+
+	return (
+		<FormControl>
+			<InputLabel htmlFor={id}>
+				{label}
+				{required ? '*' : null}
```
myarmolinsky, line 45: Not sure if I missed something but can't we use the `required` prop from `InputLabel` now that we're using that? 😕

src/components/Form/PasswordWidget.tsx
```diff
@@ -0,0 +1,126 @@
...
+				sx={[
+					STRENGTH_STYLES[strengthScore],
+					{ height: '4px', transition: 'all ease-out 150ms' },
+				]}
+			/>
+		</Box>
+	);
+};
+
+export const PasswordWidget = ({
```
myarmolinsky, line 62: nit: Might be better organized if we move the Widgets into their own folder `/src/components/Form/Widgets`, what do you think? I know we just have two right now but I plan on bringing over the FileWidget that Otavio has been working on after this PR is complete, and I won't be surprised if we find more widgets that we want to exist in this library

src/components/Form/SelectWidget.tsx
JSReds: left over from the rebase, I should have checked better

src/components/Form/Widgets/PasswordWidget.tsx
```diff
@@ -0,0 +1,126 @@
...
+	}, [password]);
+
+	if (!password || strengthScore === undefined) {
+		return null;
+	}
+
+	return (
+		<Box>
+			<Typography>
+				Password strength: <em>{STRENGTH_TITLES[strengthScore]}</em>
```
myarmolinsky, line 50: Could/should we use a Typography for the `em` as well to ensure the theme always applies to the text? (Not sure if it's necessary, just raising this in case it is. Let me know)

src/components/Form/Widgets/PasswordWidget.tsx
```diff
@@ -0,0 +1,126 @@
...
+		return onChange(value === '' ? options.emptyValue : value);
+	};
+
+	return (
+		<FormControl>
+			<TextField
+				label={label}
+				required={required}
+				id={id}
+				type={showPassword ? 'text' : 'password'}
```
myarmolinsky, line 87: If the user submits with the text field type as `text`, will their browser still offer to save the password for them? If not, then we should not be doing this and there should be a way to make this work with just `type='password'`

JSReds: @myarmolinsky it was as you said, but modern browsers and password managers often have auto-fill features that can recognise form fields based on the name attribute. 
The approch is pretty standard, the only way to prevent this behaviour today is to have autocomplete="off" which we don't use

src/components/Form/Widgets/PasswordWidget.tsx
JSReds: I would keep using native elements instead of having a component that has to declare a font familly as we discussed in the call.

PR #24, opened 2023-10-06
Add VirtualizedAutocomplete component

myarmolinsky:
Add VirtualizedAutocomplete component

Change-type: minor

### DIFF
--- a/src/components/VirtualizedAutocomplete/index.tsx
+++ b/src/components/VirtualizedAutocomplete/index.tsx
@@ -1,0 +1,277 @@
+import {
+	Autocomplete,
+	AutocompleteProps,
+	ChipTypeMap,
+	Box,
+	ListItemButton,
+	ListItemButtonProps,
+} from '@mui/material';
+import throttle from 'lodash/throttle';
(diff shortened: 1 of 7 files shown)


src/components/VirtualizedAutocomplete/index.tsx
```diff
@@ -0,0 +1,277 @@
...
+			setOptionHeightsTotal(total);
+		}
+	}, [
+		document.documentElement.clientHeight,
+		optionHeightsTotal,
+		document.getElementsByClassName('MuiAutocomplete-option'),
+	]);
+
+	return (
+		<div
```
JSReds, line 57: lets replace this with a Box for consistency like https://github.com/balena-io-modules/ui-shared-components/blob/56c77fcfaae5e1128ef8ace73beed8ddf9b08e7b/src/components/CookiesBanner/index.tsx#L108

JSReds:
Other then that I would release this first version so we can add a paginated select. We can improve it later

myarmolinsky:
LGTM

PR #27, opened 2023-11-08
Add FileWidget to RJSForm

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/components/Form/Form.stories.tsx
+++ b/src/components/Form/Form.stories.tsx
@@ -50,6 +50,11 @@ export const Default: Story = {
 					title: 'Telephone',
 					minLength: 10,
 				},
+				additionalDocuments: {
+					type: 'string',
+					format: 'data-url',
+					title: 'Additional Documents',
+				},
 			},
 		},
 		uiSchema: {
@@ -90,6 +95,15 @@ export const Default: Story = {
 			telephone: {
 				'ui:options': {
 					inputType: 'tel',
+				},
+			},
+			additionalDocuments: {
+				'ui:options': {
+					accept: {
+						'image/png': ['.png'],
+						'text/plain': ['.txt'],
+					},
+					maxSize: 13400,
 				},
 			},
 		},
--- a/src/components/Form/Widgets/FileWidget.tsx
+++ b/src/components/Form/Widgets/FileWidget.tsx
@@ -15,14 +15,20 @@ import {
 import { Input } from '@mui/base/Input';
 import type { WidgetProps } from '@rjsf/utils';
 
-export interface OnFileReadSuccessParams {
-	dataUrl: string;
-	uploadedFile: File;
-}
+// We need to forward uploadErrors onChange because RJSF does not support widget modifying
+// the default error handler.See: https://github.com/rjsf-team/react-jsonschema-form/issues/2718
+export type OnFileReadParams =
+	| {
+			dataUrl: string;
+			uploadedFile: File;
+	  }
+	| {
+			uploadErrors: string[];
+	  };
 
 type FileWidgetProps = WidgetProps & {
-	onChange: (params?: OnFileReadSuccessParams) => void;
-	value?: OnFileReadSuccessParams | string;
+	onChange: (params?: OnFileReadParams) => void;
+	value?: OnFileReadParams | string;
 };
 
 export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
@@ -36,14 +42,20 @@ export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
 	const accept = uiSchema?.['ui:options']?.accept as Accept | undefined;
 	const maxSize = uiSchema?.['ui:options']?.maxSize as number | undefined;
 
-	const dataUrl = (dataUrlValue: OnFileReadSuccessParams | string) => {
-		return typeof dataUrlValue === 'string'
-			? dataUrlValue
-			: dataUrlValue.dataUrl;
+	const dataUrl = (
+		dataUrlValue: OnFileReadParams | string,
+	): string | undefined => {
+		if (typeof dataUrlValue === 'string') {
+			return dataUrlValue;
+		}
+
+		if ('dataUrl' in dataUrlValue) {
+			return dataUrlValue.dataUrl;
+		}
 	};
 
-	const isImage = (value: OnFileReadSuccessParams | string) => {
-		return dataUrl(value).startsWith('data:image/');
+	const isImage = (value: OnFileReadParams | string) => {
+		return dataUrl(value)?.startsWith('data:image/');
 	};
 
 	const onDrop = React.useCallback(
@@ -51,19 +63,19 @@ export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
 			setErrors([]);
 
 			if (fileRejections != null && fileRejections.length > 0) {
-				const errs = fileRejections[0].errors.map(
+				const uploadErrors = fileRejections[0].errors.map(
 					({ message }) => `${fileRejections[0].file.name}: ${message}`,
 				);
-				setErrors(errs);
-				return onChange();
+				setErrors(uploadErrors);
+				return onChange({ uploadErrors });
 			}
 
 			acceptedFiles.forEach((file) => {
 				const reader = new FileReader();
 				reader.onerror = () => {
-					const err = [`Failed to upload ${file.name}`];
-					onChange();
-					setErrors(err);
+					const uploadErrors = [`Failed to upload ${file.name}`];
+					setErrors(uploadErrors);
+					onChange({ uploadErrors });
 				};
 				reader.onloadstart = () => setLoadingPercentage(0);
 				reader.onprogress = (event) => {
@@ -82,7 +94,7 @@ export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
 				reader.readAsDataURL(file);
 			});
 		}) as NonNullable<DropzoneOptions['onDrop']>,
-		[],
+		[onChange, setLoadingPercentage, setErrors],
 	);
 
 	const { getRootProps, getInputProps, isDragActive } = useDropzone({
@@ -149,47 +161,47 @@ export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
 					alignItems="center"
 					justifyContent="center"
 				>
-					<Box>
-						{value ? (
-							isImage(value) ? (
-								<Avatar
-									sx={{
-										borderRadius: 0,
-										height: '100px',
-										width: '100px',
-										backgroundSize: 'contain',
-									}}
-									src={dataUrl(value)}
-								/>
-							) : (
-								<FilePresentIcon
-									sx={{
-										fontSize: 64,
-										color: (theme) => theme.palette.grey[500],
-									}}
-								/>
-							)
-						) : (
-							<UploadFileIcon
-								sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
+					{value ? (
+						isImage(value) ? (
+							<Avatar
+								sx={{
+									borderRadius: 0,
+									height: '100px',
+									width: '100px',
+									backgroundSize: 'contain',
+								}}
+								src={dataUrl(value)}
 							/>
-						)}
-					</Box>
-					<Box>
-						<Typography>
-							<Typography variant="inherit">Drag and drop or</Typography>{' '}
-							<Typography variant="inherit">browse files</Typography>
-						</Typography>
-					</Box>
+						) : (
+							<FilePresentIcon
+								sx={{
+									fontSize: 64,
+									color: (theme) => theme.palette.grey[500],
+								}}
+							/>
+						)
+					) : (
+						<UploadFileIcon
+							sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
+						/>
+					)}
+					<Typography ml={2}>Drag and drop or browse files</Typography>
 				</Box>
 			</Box>
 
 			{errors.length !== 0 && (
-				<List>
+				<List sx={{ listStyleType: 'disc' }}>
 					{errors.map((err) => (
-						<ListItem>
+						<ListItem
+							sx={{
+								py: 0,
+							}}
+						>
 							<ListItemText
-								sx={{ m: -2, color: (theme) => theme.palette.error.main }}
+								sx={{
+									color: (theme) => theme.palette.error.main,
+									display: 'list-item',
+								}}
 							>
 								{err}
 							</ListItemText>
@@ -197,10 +209,14 @@ export const FileWidget = ({ onChange, value, ...props }: FileWidgetProps) => {
 					))}
 				</List>
 			)}
-			{value && typeof value !== 'string' && (
-				<List>
-					<ListItem>
-						<ListItemText sx={{ m: -2 }}>
+			{value && typeof value !== 'string' && 'uploadedFile' in value && (
+				<List sx={{ listStyleType: 'disc' }}>
+					<ListItem
+						sx={{
+							py: 0,
+						}}
+					>
+						<ListItemText sx={{ display: 'list-item' }}>
 							<Typography component="span" fontWeight="bold">
 								{value.uploadedFile.name}
 							</Typography>{' '}
--- a/src/components/Form/Widgets/FileWidget.tsx
+++ b/src/components/Form/Widgets/FileWidget.tsx
@@ -1,0 +1,214 @@
+import UploadFileIcon from '@mui/icons-material/UploadFile';
+import FilePresentIcon from '@mui/icons-material/FilePresent';
+import * as React from 'react';
+import { type Accept, type DropzoneOptions, useDropzone } from 'react-dropzone';
+import {
+	Box,
+	Typography,
+	CircularProgress,
+	Avatar,
+	InputLabel,
+	List,
+	ListItem,
+	ListItemText,
+} from '@mui/material';
+import { Input } from '@mui/base/Input';
+import type { WidgetProps } from '@rjsf/utils';
+
+export interface OnFileReadSuccessParams {
+	dataUrl: string;
+	uploadedFile: File;
+}
+
+type FileWidgetProps = WidgetProps & {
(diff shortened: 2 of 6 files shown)


src/components/Form/Form.stories.tsx
```diff
@@ -92,6 +97,15 @@ export const Default: Story = {
 					inputType: 'tel',
 				},
 			},
+			logoImage: {
+				'ui:options': {
+					accept: {
+						'image/png': ['.png'],
+						'text/plain': ['.txt'],
```
myarmolinsky, line 104: Why `.txt` for a logo image? 😓 
If we want to show that we can accept text documents, we should do so in another form field

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+};
+
+export const FileWidget = ({
+	onChange,
+	onError,
+	value,
+	...props
+}: FileWidgetProps) => {
+	const [isLoading, setIsLoading] = React.useState<boolean>(false);
+	const [loadingPercentage, setLoadingPercentage] = React.useState<number>(0);
```
myarmolinsky, line 36: nit: We can do this with just the `loadingPercentage` state if we make it `number | undefined` (so we set it to `undefined` when it is no longer loading and when it starts loading we set it to `0`).

I raise this because these two states are directly related to each other and it would be nice to only have to maintain one state instead

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+	const [errors, setErrors] = React.useState<string[]>([]);
+
+	const accept =
+		props.uiSchema?.['ui:options']?.accept ??
```
myarmolinsky, line 40: We should destructure all these `props.`s in the parameter's list instead of using `props.`

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+	value?: OnFileReadSuccessParams | string;
+};
+
+export const FileWidget = ({
+	onChange,
+	onError,
+	value,
+	...props
+}: FileWidgetProps) => {
+	const [isLoading, setIsLoading] = React.useState<boolean>(false);
```
myarmolinsky, line 35: FYI the `<boolean>` should not be necessary, the state should know its type because you passed a boolean for its value

myarmolinsky: Same for the `<number>` below

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+	const [loadingPercentage, setLoadingPercentage] = React.useState<number>(0);
+	const [errors, setErrors] = React.useState<string[]>([]);
+
+	const accept =
+		props.uiSchema?.['ui:options']?.accept ??
+		props.uiSchema?.value?.['ui:options']?.accept;
+	const maxSize =
+		props.uiSchema?.['ui:options']?.maxSize ??
+		props.uiSchema?.value?.['ui:options']?.maxSize;
```
myarmolinsky, line 44: What are the cases for the info we're looking for being in `value` and for it not being in `value`? Preferably only one is the correct way to pass the info, the latter doesn't sound right on first though

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+	const onDrop = React.useCallback(
+		((acceptedFiles, fileRejections) => {
+			setErrors([]);
+
+			if (fileRejections != null && fileRejections.length > 0) {
+				const errs = fileRejections[0].errors.map(
+					({ message }) => `${fileRejections[0].file.name}: ${message}`,
+				);
+				setErrors(errs);
+				onChange();
```
myarmolinsky, line 63: Why do we call `onChange` here following rejections? To unset the value if there are errors?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+				return onError?.(errs);
```
myarmolinsky, line 64: I don't see where this is used in the UI PR, could you please clarify the intent here? Why not use `enqueueSnackbar` and remove this function use? (I suspect there may be an issue with ensuring `SnackbarProvider` is used but I expect we should be able to fix it 🤔)

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: @myarmolinsky Think of it as what you do when you upload a rejected file. If you don't call onChange, the state will be the same as previously and you will still have the previous file on state but error messages displaying. so We call onChange with an empty parameter meaning "there's no file currently uploaded"

src/components/Form/Form.stories.tsx
otaviojacobi: @myarmolinsky this was for basically showing the different icon when you upload something which is not an image. What do you think of changing logoImage for "asset" which can be a .png or .txt ?

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: @myarmolinsky Yeah, this feels hacky, but it is because the AddVariable and EditVariable on the ui side pass the ui:options inside of the uiSchema.value, I am not sure why tho

src/components/Form/Form.stories.tsx
myarmolinsky: Perhaps just an additonal field for "Other documents" so that we can showcase that you can have a dropzone with limited file types permitted and one with mixed file types permitted

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			acceptedFiles.forEach((file) => {
+				const reader = new FileReader();
+				reader.onabort = () => console.log('file reading was aborted');
```
myarmolinsky, line 69: I don't think we need this

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+							dataUrl: reader.result as string,
+							uploadedFile: file,
+						});
+					}
+				};
+
+				reader.readAsDataURL(file);
+			});
+		}) as NonNullable<DropzoneOptions['onDrop']>,
+		[],
```
myarmolinsky, line 93: I think this needs to at least hold the functions we got from props

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+					flexDirection: 'column',
+					alignItems: 'center',
+					justifyContent: 'center',
+				}}
+				style={{ minHeight: '150px' }}
+				flexDirection="column"
+				alignItems="center"
+				justifyContent="center"
+				{...getRootProps()}
+				dir={undefined}
```
myarmolinsky, line 129: What's this for?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+			>
+				<input {...getInputProps()} />
```
myarmolinsky, line 131: Not sure if this should be replaced with a `<TextField>` and have the input props passed in through the appropriate props on that (called `InputProps` I think) cc @JSReds @drskullster

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+							height: '100%',
+							display: 'flex',
+							alignItems: 'center',
+							justifyContent: 'center',
+							background: 'rgba(255, 255, 255, 0.7)',
+						}}
+					>
+						<CircularProgress variant="determinate" value={loadingPercentage} />
+					</Box>
+				)}
```
myarmolinsky, line 149: I think we need to do this instead of putting the CircularProgress inside the Dropzone area and giving it a bunch of extra styling: https://mui.com/material-ui/api/circular-progress/#aria

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+							/>
+						)}
+					</Box>
+					<Box>
+						<Typography>
+							<Typography variant="inherit">Drag and drop or</Typography>{' '}
+							<Typography variant="inherit">
+								<Link>browse</Link> files
+							</Typography>
+						</Typography>
```
myarmolinsky, line 184: Is there a reason this wouldn't work? Also cc @drskullster should this text be pretending to be a `Link`? Or should we have a `plain` Button style?
```suggestion
						<Typography>Drag and drop or <Link>browse</Link> files</Typography>
```

myarmolinsky: Also wondering why there is no `onClick` for the `browse`? Isn't clicking it supposed to open the user's documents for browsing and uploading?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+							</ListItemText>
+						</ListItem>
+					))}
+				</List>
+			)}
+			{errors.length === 0 && value && typeof value !== 'string' && (
+				<List>
+					<ListItem>
+						<ListItemText sx={{ m: -2 }}>
+							<Typography component="span" variant="body1" fontWeight="bold">
```
myarmolinsky, line 206: Not sure if the `component='span'` and `variant='body1'` are both necessary, could you confirm please?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+								{err}
+							</ListItemText>
+						</ListItem>
+					))}
+				</List>
+			)}
+			{errors.length === 0 && value && typeof value !== 'string' && (
+				<List>
+					<ListItem>
+						<ListItemText sx={{ m: -2 }}>
```
myarmolinsky, line 205: Same comment for this negative margin as the above one

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+							<Typography component="span" variant="body1" fontWeight="bold">
+								{value.uploadedFile.name}
+							</Typography>{' '}
+							({value.uploadedFile.type}, {value.uploadedFile.size} bytes){' '}
```
myarmolinsky, line 209:
```suggestion
							({value.uploadedFile.type}, {value.uploadedFile.size} bytes)
```

src/components/Form/Form.stories.tsx
otaviojacobi: @myarmolinsky Not sure, I think the overall example in the storybook looks a bit odd with two Dropzone components. Maybe we can have just one with "additional documents" which accepts png/txt ?

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: @myarmolinsky I can remove the onError indeed. But IMHO the errors showing up as a list on the bottom is better than the snackbar (static for reading the error for a file in case multiple errors are present, for example)

myarmolinsky: I agree you're right, the error list at the bottom indeed looks bettwr than the snackbar, let's stick with it

src/components/Form/Form.stories.tsx
myarmolinsky: Sure, let's do that, keep it simple for now
If we want we can add a separate story to showcase the FileWidget later

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: My understanding here @myarmolinsky is that we want to use React.useCallback as documented in https://react-dropzone.js.org/ because I want to have the `isDragActive` from useDropzone().
The warning we see is a `eslint` on just because the function is not inlined and thus it can't figure it out the dependency list. However, we are not using any of the state values inside (we are just setting them) and the function does not change so I am not sure we need to pass anything on the dependency list actually

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: :man_shrugging: removing... I guess this was me trying to figure it out css

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,216 @@
...
+					</Box>
+				</Box>
+			</Box>
+
+			{errors.length !== 0 && (
+				<List>
+					{errors.map((err) => (
+						<ListItem>
+							<ListItemText
+								sx={{ m: -2, color: (theme) => theme.palette.error.main }}
```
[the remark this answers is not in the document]
otaviojacobi: @myarmolinsky Any suggestion on how to do it? by default the listItemText look very far away from each other

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: Only the span one is needed, thanks

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: @myarmolinsky I don't think this works out of the box sadly. AFAIK this is just for accessibilit (e.g. screen readers) stuff https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

I did add it anyway, but the bunch of styling for the overlay are still necessary

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: Now that is indeed an impotant question @myarmolinsky because right now the whole area is clickable (if you click anywhere it will be prompted to browse a file). I have no idea which behaviour we want in here really

src/components/Form/Widgets/FileWidget.tsx
otaviojacobi: Maybe I can just use MUI <Input /> ? https://mui.com/material-ui/api/input/

myarmolinsky: We're using TextField for all inputs, but that is in part a design decision. Would defer to Andrea and Jon in case they have another reason for preferring one or the other in this situation, but I believe the bottom line is that regardless of which input we use it will be hidden, right?

src/components/Form/Widgets/FileWidget.tsx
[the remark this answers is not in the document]
myarmolinsky: See: https://github.com/balena-io-modules/ui-shared-components/pull/27/commits/f2fcbfc45688aa18587608c90fc665258ee5dedb

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+								sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
+							/>
+						)}
+					</Box>
+					<Box>
+						<Typography>
+							<Typography variant="inherit">Drag and drop or</Typography>{' '}
+							<Typography variant="inherit">browse files</Typography>
+						</Typography>
+					</Box>
```
myarmolinsky, line 185:
```suggestion
						<Typography>
							<Typography variant="inherit">Drag and drop or</Typography>{' '}
							<Typography variant="inherit">browse files</Typography>
						</Typography>
```

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+										color: (theme) => theme.palette.grey[500],
+									}}
+								/>
+							)
+						) : (
+							<UploadFileIcon
+								sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
+							/>
+						)}
+					</Box>
```
myarmolinsky, line 179:
```suggestion
						{value ? (
							isImage(value) ? (
								<Avatar
									sx={{
										borderRadius: 0,
										height: '100px',
										width: '100px',
										backgroundSize: 'contain',
									}}
									src={dataUrl(value)}
								/>
							) : (
								<FilePresentIcon
									sx={{
										fontSize: 64,
										color: (theme) => theme.palette.grey[500],
									}}
								/>
							)
						) : (
							<UploadFileIcon
								sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
							/>
						)}
```

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,227 @@
...
+		accept,
+		multiple: false,
+		noDrag: false,
+		disabled: loadingPercentage !== undefined,
+	});
+
+	return (
+		<>
+			{schema.title && <InputLabel>{schema.title}</InputLabel>}
+			<Box
```
myarmolinsky, line 102: Note for later PR: we should probably replace these uses of `Box` with `Grid`

src/components/Form/Widgets/FileWidget.tsx
myarmolinsky: What if the `onChange` function changes along the way? Such as if, for whatever reason, we change the `onChange` function we pass into the Form while using it? Not saying we're likely to do that, just thinking we need to be prepared for it 🤔

otaviojacobi: done

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -0,0 +1,218 @@
...
+									color: (theme) => theme.palette.grey[500],
+								}}
+							/>
+						)
+					) : (
+						<UploadFileIcon
+							sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
+						/>
+					)}
+					<Typography>Drag and drop or browse files</Typography>
```
myarmolinsky, line 176:
```suggestion
					<Typography ml={2}>Drag and drop or browse files</Typography>
```

Oops, forgot to suggest this in the last suggestion, my bad

myarmolinsky (approved):

PR #29, opened 2023-11-10
Fix Autocomplete widget selection

JSReds:
Change-type: patch

### DIFF
--- a/src/components/Form/Widgets/SelectWidget.tsx
+++ b/src/components/Form/Widgets/SelectWidget.tsx
@@ -5,6 +5,7 @@ import {
 	InputLabel,
 } from '@mui/material';
 import { WidgetProps } from '@rjsf/utils';
+import { isEqual } from 'lodash';
 
 export const SelectWidget = ({
 	id,
@@ -18,10 +19,6 @@ export const SelectWidget = ({
 	options,
 	required,
 }: WidgetProps) => {
-	const change = ({ target: { value } }: any) => {
-		return onChange(value === '' ? options.emptyValue : value);
-	};
-
 	const { enumOptions, enumDisabled } = options;
 	if (!Array.isArray(enumOptions)) {
 		return null;
@@ -38,6 +35,20 @@ export const SelectWidget = ({
 		};
 	});
 
+	// 	This components needs to handle two cases, single selection and multiple selections.
+	// 	value, getOptionLabel, getOptionDisabled and onChange has to consider 2 scenarios following this type:
+	// 	option: {
+	// 	    disabled: boolean;
+	// 	    value: any;
+	// 	    label: string;
+	// 	    schema?: RJSFSchema | undefined;
+	// 	} | {
+	// 	    disabled: boolean;
+	// 	    value: any;
+	// 	    label: string;
+	// 	    schema?: RJSFSchema | undefined;
+	// 	}[]
+
 	return (
 		<FormControl>
 			<InputLabel htmlFor={id} required={required}>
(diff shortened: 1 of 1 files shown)


src/components/Form/Widgets/SelectWidget.tsx
```diff
@@ -38,6 +35,20 @@ export const SelectWidget = ({
 		};
 	});
 
+	// 	This components needs to handle two cases, single selection and multiple selections.
```
myarmolinsky, line 38: Why does it need to handle `multiple`? We do not pass the `multiple` prop to the Autocomplete
Do you mean RJSF uses SelectWidget to handle multiple too? If so then we need to find how to know that we are handling multiple instead of single and adjust accordingly

src/components/Form/Widgets/SelectWidget.tsx
```diff
@@ -58,11 +73,38 @@ export const SelectWidget = ({
...
-				getOptionDisabled={(option) => option.disabled}
+				isOptionEqualToValue={(option, value) => isEqual(option, value)}
+				getOptionLabel={(option) =>
+					Array.isArray(option)
+						? option.map((o) => o.label).join(', ')
+						: option.label
+				}
+				getOptionDisabled={(option) =>
+					Array.isArray(option)
+						? option.every((o) => o.disabled)
```
myarmolinsky, line 84: I'm confused here. To see if 1 option is disabled, we need to check if all options are disabled?

myarmolinsky: This should only apply to 1 option at a time. I don't think it's ever for multiple options, so I expect `option.disabled` to be fine 🤔

src/components/Form/Widgets/SelectWidget.tsx
```diff
@@ -58,11 +73,38 @@ export const SelectWidget = ({
...
+						: option.disabled
+				}
 				renderInput={(params) => <TextField {...params} />}
-				onChange={change}
+				onChange={(_event, selected) => {
+					if (Array.isArray(selected)) {
+						const val = selected
+							.map((item) => {
+								if (item && typeof item === 'object' && 'value' in item) {
+									return item.value;
```
myarmolinsky, line 93: I guess because we're using RJSF we can guarantee that all items have a `value`?

src/components/Form/Widgets/SelectWidget.tsx
JSReds: @myarmolinsky as explained in the comment, I mean that this case exist and the typing is considering it. I've pasted the typing to have a better view of what I mean. Todate we don't use it but we might want to use it at some point. Let's not limit the component.

src/components/Form/Widgets/SelectWidget.tsx
JSReds: I can agree on the fact that 1 is enough but option.disabled is not something you can do, please che the typing I've pested as a comment

src/components/Form/Widgets/SelectWidget.tsx
JSReds: No this is Autocomplete typing, that's the default behaviour

```
NonNullable<{
    disabled: boolean;
    value: any;
    label: string;
    schema?: RJSFSchema | undefined;
} | {
    disabled: boolean;
    value: any;
    label: string;
    schema?: RJSFSchema | undefined;
}[]> | ({
    disabled: boolean;
    value: any;
    label: string;
    schema?: RJSFSchema | undefined;
} | {
    disabled: boolean;
    value: any;
    label: string;
    schema?: RJSFSchema | undefined;
}[])[]
```

src/components/Form/Widgets/SelectWidget.tsx
myarmolinsky: > Let's not limit the component.

100% let's not limit it, just wasn't clear on whether the support was supposed to already be present or if we were introducing it here. Understood now👌

src/components/Form/Widgets/SelectWidget.tsx
myarmolinsky: Huh, gotcha

JSReds:
LGTM

PR #31, opened 2023-11-20
Hide arrow from HTML input with datalist

JSReds:
Change-type: patch

Screenshot 2023-11-20 at 17 43 40

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -458,6 +458,15 @@ export const theme = createTheme({
 						// set the fieldset > legend max-width to 100%, and this invalidate the notched: false rule for some inputs.
 						maxWidth: 0,
 					},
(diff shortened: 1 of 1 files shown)


src/theme.ts
```diff
@@ -458,6 +458,15 @@ export const theme = createTheme({
 						// set the fieldset > legend max-width to 100%, and this invalidate the notched: false rule for some inputs.
 						maxWidth: 0,
 					},
+					'[list]::-webkit-calendar-picker-indicator': {
+						display: 'none!important',
+					},
+					'[list]::-moz-calendar-picker-indicator': {
```
drskullster, line 464: not sure that we need this (no caret showing on firefox)

PR #35, opened 2023-11-21
Fix FieldTemplate layout and update react typing

JSReds:
Change-type: patch


src/components/Form/CustomTemplates/FieldTemplate.tsx
```diff
@@ -0,0 +1,83 @@
...
+		required,
+		rawErrors = [],
+		errors,
+		help,
+		description,
+		rawDescription,
+		schema,
+		uiSchema,
+		registry,
+	} = props;
```
myarmolinsky, line 40: nit: Why not just do this directly above in the parameters list? 😅

src/components/Form/CustomTemplates/FieldTemplate.tsx
```diff
@@ -0,0 +1,83 @@
...
+			) : null}
+			<FormControl
+				fullWidth={true}
+				error={rawErrors.length ? true : false}
+				required={required}
+			>
+				{children}
+				{errors}
+				{help}
+			</FormControl>
```
myarmolinsky, line 80: What happens if we have both a description and a label for the input?

src/components/Form/CustomTemplates/FieldTemplate.tsx
```diff
@@ -0,0 +1,83 @@
...
+			required={required}
+			schema={schema}
+			uiSchema={uiSchema}
+			registry={registry}
+		>
+			{displayLabel && rawDescription ? (
+				<Typography variant="caption" color="textSecondary">
+					{description}
+				</Typography>
+			) : null}
```
myarmolinsky, line 71: Each input can have a description though, right? So does every description except the highest-level one get ignored?

Also we should probably link each description to its respective input somehow, right? Just like the label is linked to the input (for accessibility and other UX)

src/components/Form/CustomTemplates/FieldTemplate.tsx
JSReds: This comment made me think that I'm actually wrong, I should not fix this here as the description cannot stay between the label and the input. The point is that since the label is the only html element that has a htmlFor attribute and it is also the only elements that should interact with the input, we should leave it as it is. 

We should apply the fix in the configVar schema instead which return something like this:

```
default: 16
description: "Define device GPU memory in megabytes."
type: "integer"
will_reboot: true
```

If we want the description to be the label of the input then we need to change description to label

JSReds: @myarmolinsky after a long thinking on it, I would actually go with this https://github.com/balena-io/balena-ui/pull/6188

PR #40, opened 2023-11-30
Set tooltip font size to match Rendition

myarmolinsky:
Set tooltip font size to match Rendition

Change-type: patch


myarmolinsky:
LGTM

PR #43, opened 2023-12-13
Fix DialogTitle fontSize being overriden by h6 fontSize

myarmolinsky:
Fix DialogTitle fontSize being overriden by h6 fontSize

Change-type: patch


JSReds (approved):

JSReds:
Not merging just to ask if we shouldn't maybe a choose a default variant to have consistency in all dialogs, like h4 or similar

drskullster:
> not merging just to ask if we shouldn't maybe a choose a default variant to have consistency in all dialogs, like h4 or similar

@JSReds consistency will come, for now we shouldn't spend too much time. I'll go back to this when I implement the new typography system in the UI :)

PR #44, opened 2023-12-18
Add missing ref on RJSForm and missing Grid system on ObjectField uiSchema

JSReds:
Change-type: patch

### DIFF
--- a/src/components/Form/Widgets/SelectWidget.tsx
+++ b/src/components/Form/Widgets/SelectWidget.tsx
@@ -18,7 +18,9 @@ export const SelectWidget = ({
 	multiple,
 	options,
 	required,
+	uiSchema,
 }: WidgetProps) => {
+	const { inputProps, ...otherUiSchema } = uiSchema ?? {};
 	const { enumOptions, enumDisabled } = options;
 	if (!Array.isArray(enumOptions)) {
 		return null;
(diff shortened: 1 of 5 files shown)


src/components/Form/Widgets/SelectWidget.tsx
```diff
@@ -18,7 +18,9 @@ export const SelectWidget = ({
 	multiple,
 	options,
 	required,
+	uiSchema,
 }: WidgetProps) => {
+	const { inputProps, ...otherUiSchema } = uiSchema ?? {};
```
myarmolinsky, line 23: Perhaps we should pass `InputProps` instead and do `<TextField {...params} {...InputProps} />` for `renderInput` instead. I think there's some props we may want to change outside of the inner `inputProps` 🤔 Not certain

JSReds: @myarmolinsky not getting the comment sorry, could you please expand on what you think the solution could be ?

just bear in mind that:
- `params` also contains some inputProps that we don't want override
- inputProps and InputProps are different props https://mui.com/material-ui/api/text-field/#TextField-prop-inputProps

JSReds:
LGTM

myarmolinsky (approved):

PR #47, opened 2024-01-09
Add getBoundingContainerRect util

JSReds:
Change-type: minor

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -1,0 +1,28 @@
+export type PartialDomRect = Partial<Omit<DOMRect, 'toJSON'>>;
+
+export const getBoundingContainerRect = (
+	element: EventTarget & HTMLDivElement,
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -0,0 +1,26 @@
...
+	const left = element.offsetLeft + parent.offsetLeft;
+	const height = element.clientHeight;
+	if (!parent || !top || !left || !height) {
+		return null;
+	}
+	return {
+		top,
+		bottom: top + height,
+		left,
+		right: left + element.clientWidth,
```
myarmolinsky, line 20: Why do we need to check `element.clientHeight` but not `element.clientWidth`?

myarmolinsky:
LGTM

src/utils/index.ts
JSReds: nice catch

PR #49, opened 2024-01-10
MUI Grid: Set the default value for `spacing` to be 2

myarmolinsky:
MUI Grid: Set the default value for `spacing` to be 2

Connects-to: https://github.com/balena-io/balena-ui/pull/6372
Change-type: patch


drskullster:
LGTM

PR #53, opened 2024-01-10
Add dynamic import on zxcvbn

JSReds:
Change-type: patch

### DIFF
--- a/src/components/Form/Widgets/PasswordWidget.tsx
+++ b/src/components/Form/Widgets/PasswordWidget.tsx
@@ -1,65 +1,22 @@
 import {
-	Box,
 	FormControl,
 	IconButton,
 	InputAdornment,
 	TextField,
-	Typography,
 } from '@mui/material';
-import * as React from 'react';
+import { Suspense, lazy, useState } from 'react';
 import { Visibility, VisibilityOff } from '@mui/icons-material';
 import { WidgetProps } from '@rjsf/utils';
-
-const STRENGTH_TITLES = ['Very weak', 'Weak', 'Good', 'Strong', 'Very strong'];
-const STRENGTH_STYLES = [
-	{ width: 0 },
-	{ width: '25%', backgroundColor: 'orange' },
-	{ width: '50%', backgroundColor: 'yellow' },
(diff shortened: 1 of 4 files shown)


tsconfig.esm5.json
```diff
@@ -1,7 +1,7 @@
 {
 	"extends": "./tsconfig.json",
 	"compilerOptions": {
-		"module": "es2015",
+		"module": "es2020",
```
JSReds, line 4: needed to support dynamic imports, IMHO no much sense to keep es2015 anymore as all our UI projects does not support old browser versions anymore:

https://github.com/balena-io/balena-ui/blob/d05abdcabcc0fc1d4cd41636a305d7ab84fface8/tsconfig.json#L3

src/components/Form/Widgets/PasswordWidget.tsx
```diff
@@ -25,19 +24,28 @@ export interface PasswordStrengthProps {
...
+		// updating the state of an unmounted component which would lead to memory leaks, as it can prevent React from properly cleaning up the component.
+		let isMounted = true;
+
+		(async () => {
 			try {
-				const { score } = zxcvbn(password);
-				setStrengthScore(score);
-			} catch {
-				// Ignore any errors, as we only want to show the strength meter if it is available.
+				const zxcvbn = (await import('zxcvbn')).default;
```
thgreasi, line 36: We probably only have to import it only the first time that the password isn't empty right? (I realize it shouldn't make much difference, but in any case it would be 1 less function call).

thgreasi: How about using 1 useEffect (w/ a useRef?) to load `zxcvbn` the first time the password has a value and this second useEffect just for the calculation?

tsconfig.esm5.json
thgreasi: I agree, but this should become a major just to be picky. We could even maybe decide to start expiring only in es2020 from now on.
Perhaps it's an opportunity to bump the `target` to at least `es2015`, like in the UI.

PR #58, opened 2024-02-13
Export `@balena/design-tokens` as `DesignTokens`

myarmolinsky:
Export `@balena/design-tokens` as `DesignTokens`

Change-type: minor


src/index.ts
```diff
@@ -67,3 +67,4 @@ export * as IconsMaterial from '@mui/icons-material';
 export { enqueueSnackbar, closeSnackbar } from 'notistack';
 export { getBoundingContainerRect } from './utils';
 export type { PartialDomRect } from './utils';
+export * as DesignTokens from '@balena/design-tokens';
```
drskullster, line 70: I would use camelCase as we're not exporting a component but a JS object.

```suggestion
export * as designTokens from '@balena/design-tokens';
```

drskullster (approved):

PR #59, opened 2024-02-15
Add `Img` component

myarmolinsky:
Add `Img` component

Change-type: minor


src/components/Img/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+	const [src, setSrc] = useState(srcProp);
+
+	return (
+		<Box
+			component="img"
+			src={src}
+			onerror={() => {
+				setSrc(fallback);
+			}}
+			{...props}
```
thgreasi, line 22:
```suggestion
			{...props}
			onerror={(err) => {
				setSrc(fallback);
				props?.onerror(err);
			}}
			
```

src/components/Img/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * This component will display an image given an src.
+ * The `fallback` prop has been added to make it easier to update the image in case the primary src fails.
+ */
+export const Img = ({
+	src: srcProp,
+	fallback,
+	...props
+}: BoxProps<'img'> & { fallback?: BoxProps<'img'>['src'] }) => {
```
JSReds, line 12: I would move in a interface for consistency and to not have to open another PR the day we will need img props

src/components/Img/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface ImgProps extends BoxProps<'img'> {
+	fallback?: BoxProps<'img'>['src'];
+}
+
+/**
+ * This component will display an image given an src.
+ * The `fallback` prop has been added to make it easier to update the image in case the primary src fails.
+ */
+export const Img = ({ src: srcProp, fallback, ...props }: ImgProps) => {
```
JSReds, line 12:
```suggestion
	src: srcProp,
	fallback,
	onerror,
	...props
```

```
onerror={() => {
				if (fallback != null) {
					setSrc(fallback);
				}
				onerror?.()
			}}
```

JSReds: sorry the diff is not right, in any case, what about desctructuring onerror and passing it to the onerror function ? just in case we need to do something else.

PR #63, opened 2024-02-19
Update typescript compilation workflow

drskullster:
* Publish only as esm (see [discussion in Zulip](https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/Buttons.20migration/near/421210310))
* Prevent storybook files from being published
* Bump ts compiler target to 2015 (matches `balena-ui` target)

The two last items reduced the dist folder size from 616K to 368K.

Change-type: major


tsconfig.json
```diff
@@ -14,13 +16,13 @@
...
 		"isolatedModules": true,
 		"noEmit": false,
 		"jsx": "react-jsx",
-		"downlevelIteration": true
+		"downlevelIteration": true,
+		"importHelpers": true
 	},
-	"include": ["src"]
+	"include": ["src"],
+	"exclude": ["src/**/*.stories.tsx"]
```
thgreasi, line 27: This exclude will now stop tsc from actually type checking the stories as well.
I think we need to create a separate tsconfig.dev.json which includes stories as well and have that run from a separate tsc statement in the tests w/ --noEmit. We will anyway have to do this once we move to the latest eslint-based balena-lint, b/c it is only able to lint files pointed by the tsconfig.

Eg:
See: https://github.com/balena-io/balena-sdk/blob/1f4b43a6a15231b090fc3c85be5eed69687408a0/package.json#L44

drskullster: nice catch @thgreasi. I can't find an easy way to point the storybook server to another file than `tsconfig.json` though, what do you think about preventing the files from getting published in a .npmignore file and removing the `exclude` rule here?

tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
 	"compilerOptions": {
-		"target": "es5",
+		"module": "esnext",
```
JSReds, line 3:
```suggestion
		"module": "es2020",
```

thgreasi (approved):
Disabled auto-merge just to discuss whether we should also bump the target to something above 2015, but feel free to get this in whenever you want.

PR #70, opened 2024-02-28
Drop support for non-es2017 compliant browsers

drskullster:
Change-type: major


JSReds:
Just for reference `es2017` browser compatibilities:



Screenshot 2024-02-28 at 16 57 09

PR #72, opened 2024-03-01
Adapt typography styling

drskullster:
Change-type: minor

### DIFF
--- a/src/components/Copy/index.tsx
+++ b/src/components/Copy/index.tsx
@@ -1,0 +1,142 @@
+import { faCopy } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Box, Tooltip } from '@mui/material';
+import React, { useEffect, useState } from 'react';
+
+export interface CopyProps {
+	copy: string | undefined;
+	onClick?: () => void;
+}
+
+export interface CopyPropsWithChildren extends CopyProps {
+	children: React.ReactNode;
+	variant?: 'absolute' | 'adjacent';
+}
+
+export const copyToClipboard = async (text: string): Promise<void> => {
+	try {
+		await navigator.clipboard.writeText(text);
+	} catch (err) {
+		console.error('Failed to copy text: ', err);
+	}
+};
+
+export const Copy = ({
+	copy,
+	onClick,
+	...props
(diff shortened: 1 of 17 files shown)


src/components/Copy/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+		console.error('Failed to copy text: ', err);
+	}
+};
+
+export const Copy = ({
+	copy,
+	onClick,
+	...props
+}: CopyProps | CopyPropsWithChildren) => {
+	let closeTooltipTimeout: ReturnType<typeof setInterval> | null = null;
```
JSReds, line 29: this cannot be a variable it must be a state, if the component re-render you will loose the information

src/components/Copy/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const copyClick = React.useCallback(
+		(e: React.MouseEvent) => {
+			e.preventDefault();
+			e.stopPropagation();
+			handleTooltipOpen();
+			onClick?.();
+			// TODO: Just to make TS happy. This will never happen because we return null if there is no copy prop
+			if (copy) {
+				copyToClipboard(copy);
+			}
```
JSReds, line 54: let's just do 

if (!copy) {
 return; 
}

on top so we also avoid handleTooltipOpen and onClick for consistency

src/components/Copy/index.tsx
JSReds: eg:


```
const [tooltipVisible, setTooltipVisible] = useState(false);
  const [closeTooltipTimeout, setCloseTooltipTimeout] = useState(null);

  // Function to show the tooltip
  const showTooltip = () => {
    setTooltipVisible(true);

    // Set timeout to automatically hide tooltip after 5 seconds
    const timeoutId = setTimeout(() => {
      setTooltipVisible(false);
      setCloseTooltipTimeout(null); // Reset timeout state after firing
    }, 5000);

    setCloseTooltipTimeout(timeoutId);
  };

  // Effect to clear timeout if closeTooltipTimeout changes or component unmounts
  useEffect(() => {
    return () => {
      if (closeTooltipTimeout) {
        clearTimeout(closeTooltipTimeout);
      }
    };
  }, [closeTooltipTimeout]);

```

JSReds (approved):

PR #76, opened 2024-03-18
Improve icon styles

drskullster:
Fix icon size too large (went from 14px to 20px after the buttons migration) and improve Icon Button styles.

Change-type: patch

### Before

image

### After

image

### Before

image

### After

image

### DIFF
--- a/src/components/DialogWithCloseButton/index.tsx
+++ b/src/components/DialogWithCloseButton/index.tsx
@@ -1,5 +1,6 @@
 import { Dialog, DialogProps, IconButton, DialogTitle } from '@mui/material';
-import CloseIcon from '@mui/icons-material/Close';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faClose } from '@fortawesome/free-solid-svg-icons/faClose';
 
 export type DialogWithCloseButtonProps = Omit<DialogProps, 'title'> & {
 	title: DialogProps['title'] | JSX.Element;
@@ -18,7 +19,7 @@ export const DialogWithCloseButton = ({
 					sx={{
 						display: 'flex',
 						justifyContent: 'space-between',
-						alignItems: 'flex-start',
+						alignItems: 'center',
 					}}
 					onClick={(e) => {
 						e.preventDefault();
@@ -29,15 +30,13 @@ export const DialogWithCloseButton = ({
 					{title}
 					{onClose ? (
 						<IconButton
+							edge="end"
 							aria-label="close"
 							onClick={(e) => {
 								onClose(e, 'backdropClick');
 							}}
-							sx={{
-								color: (theme) => theme.palette.grey[500],
-							}}
 						>
-							<CloseIcon />
+							<FontAwesomeIcon icon={faClose} />
 						</IconButton>
 					) : null}
 				</DialogTitle>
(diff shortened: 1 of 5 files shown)


src/components/DialogWithCloseButton/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { Dialog, DialogProps, IconButton, DialogTitle } from '@mui/material';
-import CloseIcon from '@mui/icons-material/Close';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faClose } from '@fortawesome/free-solid-svg-icons/faClose';
```
JSReds, line 3:
```suggestion
import faClose from '@fortawesome/free-solid-svg-icons/faClose';
```

JSReds: OR

```suggestion
import { faClose } from '@fortawesome/free-solid-svg-icons';
```

drskullster: @JSReds thanks, I'll try that. FYI I picked the import syntax from FA docs since they seem to say it's better for tree shaking: https://docs.fontawesome.com/web/use-with/react/add-icons

src/theme.ts
```diff
@@ -596,6 +602,37 @@ export const theme = createTheme({
...
+					},
+				},
+				sizeMedium: {
+					padding: '12px',
+					fontSize: '14px',
+				},
+				sizeSmall: {
+					padding: '10px',
+					fontSize: '12px',
+				},
```
myarmolinsky, line 631: nit: mind if we swap these two so that it's in order by size? 😅

src/theme.ts
```diff
@@ -596,6 +602,37 @@ export const theme = createTheme({
...
+				startIcon: {
+					' > :nth-of-type(1)': {
+						fontSize: '14px',
+					},
+				},
+				endIcon: {
+					' > :nth-of-type(1)': {
+						fontSize: '14px',
+					},
+				},
```
myarmolinsky, line 614: Why aren't we applying the font size to all icons regardless of location in the button? Not saying we should, just trying to understand why

src/theme.ts
```diff
@@ -326,6 +329,9 @@ export const theme = createTheme({
 					backgroundColor: color.bg.danger.value,
 					color: color.text.danger.value,
 				},
+				action: {
```
myarmolinsky, line 332: This is the `X`?

src/theme.ts
drskullster: I don't think I can target the element easily @myarmolinsky, there is no general `.MuiButton-icon` class. Or am I missing something?

src/theme.ts
drskullster: it is @myarmolinsky

src/theme.ts
drskullster: > I just realized actually, why do we need `> :nth-of-type` at all? Shouldn't we only ever have 1 icon in `startIcon` or `endIcon`?

I'm following MUI's CSS rule, I figure there's a good reason they're targeting it like that

drskullster: > I would expect the icon font size to just match the font size for the rest of the button tbh. Wouldn't make sense to have a 14px icon when our button text is 20px 😕

although it's currently the case this is not necessarily true @myarmolinsky, there are valid reasons to have smaller icon sizes than the text (for example if we switch to a library with sharper/bolder icons). That's why I created different tokens for icon sizes (see the typography PR).

I'll revisit this bit though, I agree the duplication of the rule is not ideal :)

myarmolinsky (approved):

PR #77, opened 2024-03-18
Add collection summary component

JSReds:
Change-type: minor

### DIFF
--- a/src/components/CollectionSummary/index.tsx
+++ b/src/components/CollectionSummary/index.tsx
@@ -1,0 +1,67 @@
+import { color } from '@balena/design-tokens';
+import { faCaretDown } from '@fortawesome/free-solid-svg-icons/faCaretDown';
+import { faCaretUp } from '@fortawesome/free-solid-svg-icons/faCaretUp';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Box, Button, Fade, Typography, Stack } from '@mui/material';
+import { useMemo, useState } from 'react';
+
+export interface CollectionSummaryProps {
+	items: string[];
+	itemsType: string;
+	expanded?: boolean;
+	maxVisibleItemCount?: number;
+}
+
+export const CollectionSummary = ({
+	expanded,
+	items,
+	itemsType,
+	maxVisibleItemCount,
+}: CollectionSummaryProps) => {
+	const [isExpanded, setIsExpanded] = useState(expanded ?? false);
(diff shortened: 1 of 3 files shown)


src/components/CollectionSummary/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+						variant="text"
+						color="secondary"
+						endIcon={isExpanded ? <ArrowDropUp /> : <ArrowDropDown />}
+						size="small"
+					>
+						{`${items.length} ${itemsType}`}
+					</Button>
+
+					<Fade in={isExpanded}>
+						<Typography variant="body2" color="text.secondary">
```
myarmolinsky, line 52: `body2` is not one of our variants, @drskullster which variant would you say is most appropriate here? Maybe `caption` or `subtitle`? (I don't remember if those are ours) Maybe `bodySm`?

src/components/CollectionSummary/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+	}, [items, maxVisibleItemCount]);
+
+	if (!items.length) {
+		return null;
+	}
+
+	return (
+		<Box>
+			{items.length === 1 ? (
+				<Typography variant="body1" color="text.secondary">
```
myarmolinsky, line 36: Same as the other comment

src/components/CollectionSummary/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { ArrowDropDown, ArrowDropUp } from '@mui/icons-material';
```
myarmolinsky, line 1: We discussed we should use free FA icons instead of `@mui/icons-material`, right?

src/components/CollectionSummary/index.tsx
drskullster: for this one I would use `body`, which is the default variant so no need to specify it :)

src/components/CollectionSummary/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+	if (!items.length) {
+		return null;
+	}
+
+	return (
+		<Box>
+			{items.length === 1 ? (
+				<Typography color="text.secondary">{stringifyItems}</Typography>
+			) : (
+				<Box display="flex" flexDirection="column" alignItems="start">
```
myarmolinsky, line 40:
```suggestion
				<Stack alignItems="start">
```

myarmolinsky (approved):

PR #82, opened 2024-04-03
Fix text button styling

drskullster:
Remove paddings around buttons with the variant `text`, disable the background color on hover and the ripple effect on active.

Change-type: patch


myarmolinsky (approved):

drskullster:
LGTM

PR #85, opened 2024-04-05
Adapt tooltip typing

JSReds:
Change-type: patch

### DIFF
--- a/src/components/DropDownButton/index.tsx
+++ b/src/components/DropDownButton/index.tsx
@@ -23,7 +23,7 @@ import { Tooltip } from '../Tooltip';
 
 type MenuItemType<T> = MenuItemWithTrackingProps &
 	T & {
-		tooltip?: string | Omit<TooltipProps, 'children'>;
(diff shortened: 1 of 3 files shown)


src/components/DropDownButton/index.tsx
```diff
@@ -23,7 +23,7 @@ import { Tooltip } from '../Tooltip';
 
 type MenuItemType<T> = MenuItemWithTrackingProps &
 	T & {
-		tooltip?: string | Omit<TooltipProps, 'children'>;
+		tooltip?: string | Omit<TooltipProps, 'children'> | null;
```
myarmolinsky, line 26:
```suggestion
		tooltip?: Omit<TooltipProps, 'children'> | TooltipProps['title'];
```

JSReds: 🤔 because TooltipProps['title'] does not have null:

`title?: string | undefined;`

myarmolinsky (approved):

PR #89, opened 2024-04-16
Add missing exports

JSReds:
Changte-type: patch


myarmolinsky:
I would just make this a `minor` since components are being made available

JSReds:
LGTM

PR #94, opened 2024-04-19
Fix AnalyticsContextProvider update state on prop change

JSReds:
Change-type: patch

### DIFF
--- a/src/contexts/AnalyticsContext.tsx
+++ b/src/contexts/AnalyticsContext.tsx
@@ -4,7 +4,7 @@ import {
 	createWebTracker,
 	WebTracker,
 } from 'analytics-client';
-import { createContext, useContext, useReducer } from 'react';
+import { createContext, useContext, useEffect, useReducer } from 'react';
 
 export enum AnalyticsStoreActions {
 	setAnalyticsData = 'set_analytics_data',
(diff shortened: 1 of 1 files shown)


src/contexts/AnalyticsContext.tsx
```diff
@@ -101,10 +106,15 @@ interface AnalyticsContextProviderProps {
...
-	);
+	const [state, dispatch] = useReducer(contextReducer, initialContext);
+	useEffect(() => {
+		if (initialState) {
+			dispatch({
+				type: AnalyticsStoreActions.setAnalyticsData,
+				payload: initialState,
+			});
+		}
+	}, [initialState]);
```
myarmolinsky, line 117: Shouldn't this useEffect just be on mount since this is initial state? I expect that if it somehow changes at any point, we still wouldn't want to suddenly change state

JSReds: nope for mainly 2 reasons: 
-  we want to override the state in any case as initialState is of type AnalyticsContext, so if for some reason the Application that use the context wants to change it, it can.

- initialState is probably null at first render, because the flow is usually as follow:

Application first render (e.g UI) => analytics client null => initialise Analytics client => Analytics client defined

myarmolinsky (approved):

PR #97, opened 2024-04-28
Export FileWidget as component rather than type

otaviojacobi:
Change-type: patch

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.8.2",
       "license": "ISC",
       "dependencies": {
-        "@balena/design-tokens": "^0.7.2",
+        "@balena/design-tokens": "^0.8.0",
         "@emotion/react": "^11.10.6",
         "@emotion/styled": "^11.10.6",
         "@fortawesome/free-solid-svg-icons": "^6.5.1",
@@ -2222,9 +2222,9 @@
       }
     },
     "node_modules/@balena/design-tokens": {
-      "version": "0.7.2",
-      "resolved": "https://registry.npmjs.org/@balena/design-tokens/-/design-tokens-0.7.2.tgz",
-      "integrity": "sha512-vV6CqTVfxUN8ZmZ6jrvfCfugOSmo/cBwZRrbFMSzTb9g/rQzC90kN3keY6aeLTbkgJhcrljqeeaJTTi5ckV45Q=="
+      "version": "0.8.0",
+      "resolved": "https://registry.npmjs.org/@balena/design-tokens/-/design-tokens-0.8.0.tgz",
+      "integrity": "sha512-QUYjdjR1ahTN98NpH3KJ00gd7u42v88IdkIbpjqTY8QaMZVWgdF5N0WArya91cV581zvKw+Lv7Y6JHg9c6wSPw=="
     },
     "node_modules/@balena/lint": {
       "version": "6.2.2",
--- a/src/components/Form/Widgets/FileWidget.tsx
+++ b/src/components/Form/Widgets/FileWidget.tsx
@@ -1,248 +1,476 @@
-import UploadFileIcon from '@mui/icons-material/UploadFile';
-import FilePresentIcon from '@mui/icons-material/FilePresent';
-import { type Accept, type DropzoneOptions, useDropzone } from 'react-dropzone';
 import {
-	Box,
+	type Accept,
+	type DropzoneOptions,
+	useDropzone,
+	ErrorCode,
+	FileRejection,
+} from 'react-dropzone';
+import {
 	Typography,
-	CircularProgress,
-	Avatar,
 	InputLabel,
-	List,
-	ListItem,
-	ListItemText,
+	Stack,
+	type SxProps,
+	Link,
+	LinearProgress,
+	Box,
+	useMediaQuery,
+	useTheme,
 } from '@mui/material';
 import type { WidgetProps } from '@rjsf/utils';
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
+import { IconButtonWithTracking, Tooltip, designTokens } from '../../..';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import {
+	IconDefinition,
+	faCircleExclamation,
+	faFile,
+	faFileImage,
+	faTrash,
+	faUpload,
+} from '@fortawesome/free-solid-svg-icons';
+import uniq from 'lodash/uniq';
 
-export interface OnFileReadSuccessParams {
+const restingStyle: SxProps = {
+	borderColor: designTokens.color.border.value,
+	borderWidth: '1px',
+	backgroundColor: designTokens.color.bg.value,
+};
+
+const hoverStyle: SxProps = {
+	borderColor: designTokens.color.border.strong.value,
+	borderWidth: '1px',
+	backgroundColor: designTokens.color.bg.value,
+	'.browse-files-text': { textDecoration: 'none' },
+};
+
+const disabledStyle: SxProps = {
+	borderColor: designTokens.color.border.strong.value,
+	borderWidth: '1px',
+	backgroundColor: designTokens.color.bg.value,
+	opacity: '40%',
+};
+
+const dragStyle: SxProps = {
+	borderColor: designTokens.color.border.accent.strong.value,
+	borderWidth: '2px',
+	backgroundColor: designTokens.color.bg.accent.value,
+};
+
+// These are just the ones we use in the dashboard at the moment
+// (excluding XML which is a pro icon, for which we just default to faFile)
+// we can add more as we get the usecases
+// TODO: Add the XML pro icon when we have support for pro icons in this repo
+const getFileTypeIcon = (fileType: string) => {
+	switch (fileType) {
+		case 'image/png':
+		case 'image/jpg':
+		case 'image/jpeg':
+			return faFileImage;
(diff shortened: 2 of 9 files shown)


src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -204,9 +207,11 @@ export const FileWidget = ({
 							sx={{ fontSize: 64, color: (theme) => theme.palette.grey[500] }}
 						/>
 					)}
-					<Typography ml={2}>Drag and drop or browse files</Typography>
-				</Box>
-			</Box>
+					<Typography ml={2}>
```
myarmolinsky, line 210: remove `ml={2}` for `gap={2}` on the Stack

otaviojacobi: Are you doing it or do you want me to push :P ?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,429 @@
...
-
-		if ('dataUrl' in dataUrlValue) {
-			return dataUrlValue.dataUrl;
-		}
-	};
-
-	const isImage = (value: OnFileReadParams | string) => {
-		return dataUrl(value)?.startsWith('data:image/');
-	};
+	const disabled = uiSchema?.['ui:options']?.disabled as boolean | undefined;
```
JSReds, line 115: To ensure clarity, should we add a `disabledReasons` property and a tooltip to explain to the user why drag-and-drop is disabled?

Also, the cast here is not necessary

```suggestion
	const disabled = uiSchema?.['ui:options']?.disabled;
```

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,429 @@
...
+			if (multiple) {
+				setFiles([
+					...files,
+					...acceptedFiles,
+					...rejectedFiles.map(
+						(rejectedFile) =>
+							({
+								...rejectedFile.file,
+								// for some reason, spreading rejectedFile.file does not spread these properties
+								name: rejectedFile.file.name,
```
JSReds, line 192: It's expected as the interfaces are different.

```
type AcceptedFile = File & {
    loadingPercentage?: number;
    base64Data?: string;
    errors?: Array<{ code: ErrorCode | undefined; message: string }>;
};

interface FileRejection {
    file: File;
    errors: FileError[];
}
```

The issue isn't with the spread operator failing to spread properties;
rather, these properties simply don't exist on the object.

Additionally, why are we attempting to display the file if it was rejected?
If the file is rejected, I would expect the widget not load the file but instead only display the error.

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,429 @@
...
+			} else {
+				setFiles(
+					!!acceptedFiles.length
+						? [acceptedFiles[0]]
+						: !!rejectedFiles.length
+						? [
+								{
+									...rejectedFiles[0].file,
+									// for some reason, spreading rejectedFile.file does not spread these properties
+									name: rejectedFiles[0].file.name,
```
JSReds, line 208: expected as above

src/index.tsx
```diff
@@ -24,12 +24,7 @@ export type {
 	UiSchema,
 	FormValidation,
 } from './components/Form';
-export type {
-	FileWidget,
```
JSReds, line 28: I recommend keeping the export on the fileWidget, as it might be needed in the future.

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,429 @@
...
+};
+
+const dragStyle: SxProps = {
+	borderColor: designTokens.color.border.accent.strong.value,
+	borderWidth: '3px',
+	backgroundColor: designTokens.color.bg.accent.value,
+};
+
+// These are just the ones we use in the dashboard at the moment
+// (excluding XML which is a pro icon, for which we just default to faFile)
```
drskullster, line 49: Maybe add a TODO to add the XML icon once we have pro icons in ui-shared?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,429 @@
...
+		case 'image/jpg':
+		case 'image/jpeg':
+			return faFileImage;
+		default:
+			return faFile;
+	}
+};
+
+const fileSizeToReadable = (size: number) => {
+	const i = Math.floor(Math.log(size) / Math.log(1000));
```
drskullster, line 63: Shouldn't it be `1024`? 🤔

src/components/Form/Widgets/FileWidget.tsx
myarmolinsky: The idea is to show the user a list of the files they tried to upload and the reasons the uploads failed per file. I will send the design in zulip, seeing it should be helpful

src/components/Form/Widgets/FileWidget.tsx
myarmolinsky: 1024 is binary, I'm pretty sure react dropzone uses decimal

otaviojacobi: I also think it should be 1024. This is still decimal, just breaking up on 1024bytes at a time

package-lock.json
```diff
@@ -9,7 +9,7 @@
       "version": "5.8.2",
       "license": "ISC",
       "dependencies": {
-        "@balena/design-tokens": "^0.7.2",
```
JSReds, line 12: This file contains numerous changes that are unrelated to the module installation. Could you please install only the specific module without running npm install?

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,478 @@
...
+		'small';
+	const uploadIcon = uiSchema?.['ui:options']?.uploadIcon as
+		| IconDefinition
+		| undefined;
+	const descriptiveText = uiSchema?.['ui:options']?.descriptiveText as
+		| string
+		| undefined;
+	const hideUploadedFiles = uiSchema?.['ui:options']?.hideUploadedFiles as
+		| boolean
+		| undefined;
```
JSReds, line 147: what about something like this to avoid all these type casts ? (feel free to update it, just a basic idea)

```
type UIOptions = {
	accept?: Accept;
	maxSize?: number;
	disabled?: boolean;
	disabledReason?: string;
	browseResourceName?: string;
	variant?: 'large' | 'small';
	uploadIcon?: IconDefinition;
	descriptiveText?: string;
	hideUploadedFiles?: boolean;
};

const options = uiSchema?.['ui:options'] as UIOptions | undefined;

const accept = options?.accept;
const maxSize = options?.maxSize;
const disabled = options?.disabled;
const disabledReason = options?.disabledReason;
const browseResourceName = options?.browseResourceName;
const isSmall = options?.variant === 'small';
const uploadIcon = options?.uploadIcon;
const descriptiveText = options?.descriptiveText;
const hideUploadedFiles = options?.hideUploadedFiles;
```

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,478 @@
...
+							{errorFiles
+								.map((errorFile) => errorFile.errors)
+								.flat()
+								.map((error) =>
+									getErrorMessage(
+										error.code,
+										error.message,
+										maxSize ?? Infinity,
+									),
+								)}
```
JSReds, line 340:
```suggestion
.flatMap((errorFile) => errorFile.errors)
.map((error) =>
	getErrorMessage(
		error.code,
		error.message,
		maxSize ?? Infinity,
	),
)}
```

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,478 @@
...
+											</Typography>
+											<Typography color={designTokens.color.text.subtle.value}>
+												{fileSizeToReadable(file.size)}
+											</Typography>
+										</Stack>
+										<IconButtonWithTracking
+											eventName="FileWidget: Delete uploaded file"
+											aria-label="Delete uploaded file"
+											onClick={() => removeFile(index)}
+											// TODO: Jon please confirm
```
JSReds, line 410: I would add it as a PR comment more then a TODO

src/components/Form/Widgets/FileWidget.tsx
myarmolinsky: I was thinking about the same, don't remember why I dropped the idea, let me do it

src/components/Form/Widgets/FileWidget.tsx
```diff
@@ -1,248 +1,477 @@
...
+											</Typography>
+											<Typography color={designTokens.color.text.subtle.value}>
+												{fileSizeToReadable(file.size)}
+											</Typography>
+										</Stack>
+										<IconButtonWithTracking
+											eventName="FileWidget: Delete uploaded file"
+											aria-label="Delete uploaded file"
+											onClick={() => removeFile(index)}
+											color="secondary"
```
myarmolinsky, line 410: @drskullster please confirm we should indeed be changing the color of this button to "secondary"

drskullster: it should be the default color @myarmolinsky, but I realize it's not the same in my mockup than in the design system. I say just remove the `color` prop and roll with that, we'll adapt in a future PR if needs be.

JSReds (approved):

PR #101, opened 2024-06-20
Fix FileWidget showing half-loading progress bar for all files

myarmolinsky:
Fix FileWidget showing half-loading progress bar for all files

Change-type: patch


myarmolinsky:
I self certify!

myarmolinsky:
I self-certify!

PR #105, opened 2024-06-26
Add SettingsCard component

drskullster:
Change-type: minor

### DIFF
--- a/src/components/SettingsCard/index.tsx
+++ b/src/components/SettingsCard/index.tsx
@@ -1,0 +1,89 @@
+import React from 'react';
+import {
+	Box,
+	Card,
+	CardContent,
+	CardHeader,
+	Stack,
+	Typography,
+} from '@mui/material';
+import { color } from '@balena/design-tokens';
+
+export interface SettingsCardProps {
+	title?: string | React.ReactNode;
+	children: React.ReactNode;
+	danger?: boolean;
+}
+
+export interface SettingsCardItemProps {
+	title: string | React.ReactNode;
+	description?: string | React.ReactNode;
+	// TODO make it required once username and social accounts can be updated in the UI
+	action?: React.ReactNode;
+}
+
+const settingsCardItemTitleClass = 'settings-card-item-title';
+
+export const SettingsCard = ({
+	title,
+	children,
+	danger,
+}: SettingsCardProps) => (
+	<Card>
+		{title && (
+			<CardHeader
+				title={title}
+				{...(danger && {
+					sx: {
+						'.MuiCardHeader-title': {
+							color: color.text.danger.value,
+						},
+					},
+				})}
+			/>
+		)}
+		{/* TODO find out why there's a paddingBottom: 24px on CardContent :last-child */}
+		<CardContent
+			sx={{
+				'&:last-child': { pb: 0 },
+				gap: 3,
+				display: 'flex',
+				flexDirection: 'column',
+				...(danger && {
+					[`.${settingsCardItemTitleClass}`]: {
+						color: color.text.danger.value,
+					},
+				}),
+			}}
+		>
+			{children}
+		</CardContent>
+	</Card>
+);
+
+export const SettingsCardItem = ({
+	title,
+	description,
+	action,
+}: SettingsCardItemProps) => (
+	<Stack
+		gap={2}
+		direction={['column', 'column', 'row']}
+		justifyContent="space-between"
+		alignItems={['start', 'start', 'center']}
+		sx={{
+			'&:not(&:last-child)': {
+				borderBottom: `dashed ${color.border.subtle.value} 1px`,
+				pb: 3,
+			},
+		}}
+	>
+		<Stack gap={2} maxWidth={[undefined, undefined, '70%']}>
+			<Typography className={settingsCardItemTitleClass} variant="bodyLg">
+				{title}
+			</Typography>
+			<Typography color="text.secondary">{description}</Typography>
+		</Stack>
+		<Box>{action}</Box>
+	</Stack>
+);
--- a/src/components/SettingsCard/SettingsCard.stories.tsx
+++ b/src/components/SettingsCard/SettingsCard.stories.tsx
@@ -1,0 +1,58 @@
+import { Button, Chip, Link, Stack, Switch, Typography } from '@mui/material';
+import type { Meta, StoryObj } from '@storybook/react';
+import { SettingsCard, SettingsCardItem } from '.';
+import { Code } from '../Code';
+
+const meta = {
+	title: 'Patterns/Settings Card',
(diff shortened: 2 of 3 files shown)


src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	danger?: boolean;
+}
+
+export interface SettingsPaneItemProps {
+	title: string | React.ReactNode;
+	description?: string | React.ReactNode;
+	action?: React.ReactNode;
+}
+
+export const SettingsPane = ({
```
myarmolinsky, line 24: I don't think `Pane` is the right word. When I think "Pane" I think of something that expands, i.e. a sliding door (slide open a sidebar by clicking a collapsed version of it)

Perhaps `Settings` is enough? Or `SettingsCard`?

src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+export interface SettingsPaneProps {
+	title?: string | React.ReactNode;
+	children: React.ReactNode;
+	danger?: boolean;
+}
+
+export interface SettingsPaneItemProps {
+	title: string | React.ReactNode;
+	description?: string | React.ReactNode;
+	action?: React.ReactNode;
```
myarmolinsky, line 21: Do we have an instance where there is no action? This feels like it should be required

src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+			},
+		}}
+	>
+		<Stack gap={2} maxWidth={[undefined, undefined, '70%']}>
+			<Typography variant="bodyLg">{title}</Typography>
+			<Typography variant="body" color="text.secondary">
+				{description}
+			</Typography>
+		</Stack>
+		<Box>{action}</Box>
```
myarmolinsky, line 68: Why wrap it in a Box?

src/components/SettingsPane/SettingsPane.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+	args: {
+		title: 'Account Settings',
+		children: (
+			<>
+				<SettingsPaneItem
+					title="Username"
+					description={
+						<Stack gap={2}>
+							<Code sx={{ width: 'fit-content' }}>johndoe113</Code>
+							<Typography variant="body" component="p">
```
myarmolinsky, line 25: That's the default variant, right? 
```suggestion
							<Typography component="p">
```

src/components/SettingsPane/index.tsx
myarmolinsky: I see the username example in the story (which I assume is because we have it in the account preferences page 🤔 

If we have no other such instances of no action, I think we can keep this *for now* but please leave a TODO that says we should make `action` not optional once usernames can be changed. We have had a GH issue for a bit now (see: https://github.com/balena-io/balena-ui/issues/1150) and I think Thodoris has told me it's technically possible and just needs to be implemented

OR you can make it required from the getgo (so we don't have to do a major later) and then give username a fake action (with a TODO of course)

src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	<Card>
+		{title && (
+			<CardHeader
+				title={title}
+				sx={{
+					'.MuiCardHeader-title': {
+						color: danger ? color.text.danger.value : 'inherit',
+					},
+				}}
+			/>
```
myarmolinsky, line 38: So that we don't have to explicitly pass `inherit`
```suggestion
				{...(danger && { sx: {
					'.MuiCardHeader-title': {
						color: danger ? color.text.danger.value : 'inherit',
					},
				}})}
			/>
```

src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+		)}
+		<CardContent sx={{ '&:last-child': { pb: 0 } }}>{children}</CardContent>
```
myarmolinsky, line 40: Is this `sx` necessary? You have `'&:not(&:last-child)'` on the `SettingsPaneItem` below

src/components/SettingsPane/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+			'&:not(&:last-child)': {
+				borderBottom: `dashed ${color.border.subtle.value} 1px`,
+				pb: 3,
+				mb: 3,
+			},
+		}}
+	>
+		<Stack gap={2} maxWidth={[undefined, undefined, '70%']}>
+			<Typography variant="bodyLg">{title}</Typography>
+			<Typography variant="body" color="text.secondary">
```
myarmolinsky, line 64:
```suggestion
			<Typography color="text.secondary">
```

src/components/SettingsPane/index.tsx
drskullster: `SettingsCard` 👌

src/components/SettingsPane/index.tsx
drskullster: I'll go with optional and the TODO @myarmolinsky, if after we refactor all our settings page we end up with only those two cases (username and social account) then we can require it

src/components/SettingsPane/index.tsx
drskullster: To allow for a 70% / 30% split while keep the right button size (otherwise the button will take all 30%)

src/components/SettingsPane/index.tsx
drskullster: so I'm not sure why but we have a padding-bottom that's added to `CardContent:last-child`. This removes it (and I have to target `&:last-child` otherwise it gets overriden by the cascade).

drskullster: I looked at the theme but can't find this rule, maybe you have an idea @myarmolinsky?

drskullster: Anyway for the sake of merging this PR I'll add a TODO to find out why we need this

src/components/SettingsCard/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+		alignItems={['start', 'start', 'center']}
+		sx={{
+			'&:not(&:last-child)': {
+				borderBottom: `dashed ${color.border.subtle.value} 1px`,
+				pb: 3,
+			},
+		}}
+	>
+		<Stack gap={2} maxWidth={[undefined, undefined, '70%']}>
+			<Typography variant="bodyLg">{title}</Typography>
```
myarmolinsky, line 75: Give a CSS rule to SettingsCard to set the titles to danger color if the SettingsCard has `danger` `true`

myarmolinsky (approved):

PR #106, opened 2024-06-26
RJSF SelectWidget: Fix form thinking field is empty when `multiple` and `required`

myarmolinsky:
RJSF SelectWidget: Fix form thinking field is empty when `multiple` and `required`

Change-type: patch


myarmolinsky:
I self-certify!

PR #110, opened 2024-08-02
Restyle ToggleButtonGroup component

JSReds:
Change-type: patch

### DIFF
--- a/src/docs/30_mui_components/buttons.mdx
+++ b/src/docs/30_mui_components/buttons.mdx
@@ -118,12 +118,21 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 <ThemeProvider theme={theme}>
 	<Box sx={{ display: 'flex', gap: 2, flexWrap: 'wrap' }}>
-		<LensToggle color="standard" />
+		<LensToggle />
 		<LensToggle color="primary" />
 		<LensToggle color="secondary" />
 		<LensToggle color="info" />
 		<LensToggle color="success" />
 		<LensToggle color="warning" />
 		<LensToggle color="error" />
+	</Box>
+	<Box sx={{ display: 'flex', gap: 2, mt: 2, flexWrap: 'wrap' }}>
+		<LensToggle>Default</LensToggle>
+		<LensToggle color="primary">Primary</LensToggle>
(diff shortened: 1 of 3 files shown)


src/docs/30_mui_components/buttons.mdx
```diff
@@ -124,6 +124,13 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 		<LensToggle color="info" />
 		<LensToggle color="success" />
 		<LensToggle color="warning" />
-		<LensToggle color="error" />
+	</Box>
+	<Box sx={{ display: 'flex', gap: 2, mt: 2, flexWrap: 'wrap' }}>
+		<LensToggle color="standard">Toggle text</LensToggle>
```
myarmolinsky, line 129: The hover color for the unselected one should be `bg.strong` instead of `bg.accent`

src/docs/30_mui_components/buttons.mdx
```diff
@@ -124,6 +124,13 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 		<LensToggle color="info" />
 		<LensToggle color="success" />
 		<LensToggle color="warning" />
-		<LensToggle color="error" />
+	</Box>
```
myarmolinsky, line 127: The hover color for the first LensToggle's unselected option should be `bg.strong` instead of `bg.accent`

src/docs/30_mui_components/buttons.mdx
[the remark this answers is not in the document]
myarmolinsky: Same for the `secondary` color text-less example above ^

src/docs/30_mui_components/buttons.mdx
```diff
@@ -124,6 +124,13 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
+	<Box sx={{ display: 'flex', gap: 2, mt: 2, flexWrap: 'wrap' }}>
```
myarmolinsky, line 128: I'm not a fan of the hover for the unselected option matching the active color in many of the cases (primary, info, success, warning). Do you think maybe we can try dimming (but not turning off) the active option when hovering over an unselected option?

JSReds:
LGTM

PR #113, opened 2024-08-20
Improve Toggle Buttons

drskullster:
* Add small improvements (color, sizes,...) to toggle buttons
* Remove most color variants for toggle buttons
* Make button height more consistent (37px)
* Fix design token list in the documentation
* Improve documentation for toggle buttons

### DIFF
--- a/src/theme.tsx
+++ b/src/theme.tsx
@@ -7,7 +7,7 @@ import {
 } from '@mui/material';
 import type {} from '@mui/x-data-grid/themeAugmentation';
 import { TypographyStyleOptions } from '@mui/material/styles/createTypography';
-import { color, typography } from '@balena/design-tokens';
+import { color, typography, shape } from '@balena/design-tokens';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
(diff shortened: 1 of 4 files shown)


src/theme.tsx
```diff
@@ -849,81 +890,61 @@ export const theme = createTheme({
...
 					margin: '2px',
-					padding: '8px',
+					paddingTop: '5px', // 5px = button padding - margin - border width
+					paddingBottom: '5px',
+					paddingLeft: theme.spacing(3),
+					paddingRight: theme.spacing(3),
+					'&:has(> svg:only-child)': {
+						// Fix width when SVG is the only child
+						// FIXME this also targets buttons that have an icon + some text node.
+						//  Text needs to be wrapped in a HTML element to prevent the fixed width.
```
myarmolinsky, line 917: We can fix this with a data attribute or className, right?

drskullster: that would be ideal, but can we add that in a global way on a MUI component? That would be "if this prop is set and this prop is empty, add a class"

myarmolinsky (approved):

PR #118, opened 2024-09-25
Tabs: Use MUI's `scrollable` variant

myarmolinsky:
Tabs: Use MUI's `scrollable` variant

Change-type: minor


myarmolinsky:
I self-certify!

PR #119, opened 2024-10-03
DownloadImageDialog: Display device type alerts as Callouts

myarmolinsky:
DownloadImageDialog: Display device type alerts as Callouts

Change-type: patch


myarmolinsky:
I self-certify!

PR #128, opened 2024-10-24
DownloadImageDialog: Fix versions dropdown not displaying selection

myarmolinsky:
DownloadImageDialog: Fix versions dropdown not displaying selection

Change-type: patch


JSReds (approved):

thgreasi (commented):
A more appropriate commit message would be:
Fix support for versions with prod & dev variants

PR #130, opened 2024-10-30
Add Notification component

drskullster:
Add a notification component to start experimenting with in-app notifications in the dashboard

Fibery: https://balena.fibery.io/Work/Project/Bring-back-product-roundup-posts-928

Zulip: https://balena.zulipchat.com/#narrow/stream/408336-aspect.2Fuser-experience/topic/How.20to.20announce.20product.20updates.3F/near/445631312

Change-type: minor


src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+	onClose?: (event: React.SyntheticEvent<any>) => void;
+}
+const NotificationRoot = styled(Box, {
+	name: 'Notification',
+	slot: 'root',
+})<{ ownerState: NotificationProps }>(({ theme, ownerState }) => ({
+	alignItems: 'center',
+	backgroundColor: 'white',
+	borderRadius: shape.border_radius.md.value + 'px',
+	boxShadow: '0px 0px 8px 0px rgba(35, 68, 94, 0.3)', // TODO add a design token for box shadows
```
myarmolinsky, line 28: Couldn't we somehow use whatever MUI applies to dialogs?

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+	width: '100%',
+}));
+
+const NotificationIllustration = styled('figure', {
+	name: 'Notification',
+	slot: 'illustration',
+})(({ theme }) => ({
+	margin: 0,
+	width: 48,
+	minWidth: 48,
```
myarmolinsky, line 43: Doesn't the first make the second pointless?

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+})<{ ownerState: NotificationProps }>(({ theme, ownerState }) => ({
+	alignItems: 'center',
+	backgroundColor: 'white',
+	borderRadius: shape.border_radius.md.value + 'px',
+	boxShadow: '0px 0px 8px 0px rgba(35, 68, 94, 0.3)', // TODO add a design token for box shadows
+	display: 'flex',
+	gap: theme.spacing(3),
+	maxWidth: 400,
+	padding: theme.spacing(3),
+	position: 'relative',
```
myarmolinsky, line 33: Is this only needed for the small breakpoint of `NotificationContent`? If so we should only apply this during that breakpoint as well, right?

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+			title,
+			linkHref,
+			linkLabel,
+			onClose,
+		};
+		const { state: analytics } = useAnalyticsContext();
+
+		useEffect(() => {
+			analytics.webTracker?.track('Notification: open', { campaignId });
+		}, []);
```
myarmolinsky, line 120: We want to track when it *appears*? Feels like a lot of useless events, what is the purpose?

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+						href={linkHref}
+						eventName="Notification: click on link"
+						eventProperties={{ campaignId }}
+					>
+						{linkLabel}{' '}
+						<FontAwesomeIcon size="sm" icon={faArrowUpRightFromSquare} />
+					</MUILinkWithTracking>
+				</NotificationContent>
+				<NotificationCloseContainer>
+					<IconButtonWithTracking
```
myarmolinsky, line 140: Is the Container necessary or can we apply its styling directly to the `IconButtonWithTracking`?

src/components/Notification/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+					<MUILinkWithTracking
+						href={linkHref}
+						eventName="Notification: click on link"
+						eventProperties={{ campaignId }}
+					>
+						{linkLabel}{' '}
+						<FontAwesomeIcon size="sm" icon={faArrowUpRightFromSquare} />
+					</MUILinkWithTracking>
+				</NotificationContent>
+				<NotificationCloseContainer>
```
myarmolinsky, line 139: Should this really be taking an entire portion in the right of the Notification? Seems it would be better to have inside the NotificationTitle aligned to the far right (like what we do for the `DialogWithCloseButton`)

src/components/Notification/index.tsx
drskullster: Ideally I would use a Paper component but I don't like the default MUI shadow. I will look at it separately and come back here to remove the random value.

src/components/Notification/index.tsx
drskullster: That's a good question, my hunch is that it's better to keep it there at all time than to add more media queries to 'toggle' it.

src/components/Notification/index.tsx
drskullster: I want to make sure that the notification appears, and appears only once 😅 I'll add a todo to remove it in the future

src/components/Notification/index.tsx
drskullster: removed because of the changes in the close icon (no need for absolute positioning anymore)

src/components/Notification/index.tsx
drskullster: I changed to `maxWidth` for clarity (and reduced the min width value)

src/components/Announcement/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+			...props,
+			campaignId,
+			children,
+			illustration,
+			title,
+			linkHref,
+			linkLabel,
+			onClose,
+			open,
+		};
```
myarmolinsky, line 122: Isn't this exactly the same as `props`? Destructuring `campaignId` (as well as any of the others) does not remove it from props, it just creates a dedicated const with its value. So I assume you can remove this entire const and just pass `props` for `ownerState`

src/components/Announcement/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+		const { state: analytics } = useAnalyticsContext();
```
myarmolinsky, line 123: Should we only use this inside the useEffect that calls it? It only runs on mount and we don't use `analytics` anywhere else so I guess we don't need the const to live outside of the useEffect

src/components/Announcement/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+			<Snackbar
+				open={open}
+				anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
+				TransitionComponent={(props: SlideProps) => {
+					return <Slide direction="up" {...props} />;
+				}}
+				ref={ref}
+				onClose={onClose}
+				sx={[
+					(theme) => ({ maxWidth: 350, zIndex: theme.zIndex.drawer - 1 }),
```
myarmolinsky, line 140: Just curious, how did you settle on this specific zIndex? It may be good to leave a comment above this line so that we don't forget/lose the context

src/components/Announcement/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+					// See: https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop
+					...(Array.isArray(sx) ? sx : [sx]),
+				]}
+			>
+				<AnnouncementRoot ownerState={ownerState} {...other}>
```
myarmolinsky, line 145: Not sure if I'm missing something but what is the point of `ownerState`? It's not being used in `AnnouncementRoot`'s `styled`

src/components/Announcement/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+								sx={(theme) => ({ ml: 'auto', my: `-${theme.spacing(1)}` })}
+							>
+								<FontAwesomeIcon icon={faClose} />
+							</IconButtonWithTracking>
+						</AnnouncementTitle>
+						{children}
+						<MUILinkWithTracking
+							href={linkHref}
+							eventName="Announcement: click on link"
+							eventProperties={{ campaignId }}
```
myarmolinsky, line 171: Perhaps we should track the link itself too? Maybe the link label too but that may be a stretch :P My thinking for the latter is maybe we can see what verbiage leads to more clicks

src/components/Announcement/index.tsx
drskullster: I think you're right, though oddly it's what's recommended in the doc: https://mui.com/material-ui/customization/creating-themed-components/#3-style-the-slot-with-ownerstate

But I just removed it because we don't need `ownerState` in the component for now.

src/components/Announcement/index.tsx
drskullster: I wanted to prevent the component showing on top of the support menu. The default z-index for snackbars is actually very high, only `tooltip` is higher.

src/components/Announcement/index.tsx
drskullster: good catch, removed it :)

src/components/Announcement/index.tsx
drskullster: I think it would break the rule of hooks https://legacy.reactjs.org/warnings/invalid-hook-call-warning.html#breaking-the-rules-of-hooks

myarmolinsky: Yeah sorry I forgot this was a hook 😖

myarmolinsky (approved):

PR #132, opened 2024-11-11
Add HighlightedName component to dynamically highlight text based on input string

JSReds:
Change-type: minor


src/components/HighlightedName/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+	color,
+	...props
+}: TypographyProps) => {
+	const theme = useTheme();
+
+	if (typeof children !== 'string') {
+		throw new Error(
+			`The child element of the HighlightedName component must be a string, received: ${typeof children}`,
+		);
+	}
```
myarmolinsky, line 30: Wouldn't it be better to omit `children` from `TypographyProps` and add our own `& { children: string }`?

myarmolinsky (approved):

PR #151, opened 2024-11-18
Bump lint

myarmolinsky:


src/components/VirtualizedAutocomplete/VirtualizedAutocomplete.stories.tsx
```diff
@@ -25,69 +24,28 @@ export const Default: VirtualizedStory = {
...
-					const filteredArray = paginatedArray.filter((option) =>
-						JSON.stringify(option).includes(query ?? ''),
-					);
-					return { data: filteredArray, totalItems: filteredArray.length + 1 };
-				}
-				return { data: paginatedArray, totalItems: paginatedArray.length + 1 };
-			}}
-		/>
-	);
-};
```
JSReds, line 82: wouldn't it be better to keep the TODO and analyse why this is not working when we have time ?

JSReds (approved):

PR #152, opened 2024-11-18
Improvements to the announcement notification

drskullster:
* Close the announcement when user has clicked on a link
* Announcement slides down when closed
* Better responsive styles
* Make room for the 'need help' button

Change-type: patch

### DIFF
--- a/src/components/Announcement/index.tsx
+++ b/src/components/Announcement/index.tsx
@@ -19,7 +19,10 @@ import { useEffect } from 'react';
 import { useAnalyticsContext } from '../../contexts/AnalyticsContext';
 import { SnackbarCloseReason } from '@mui/material/Snackbar/Snackbar';
 
-export type AnnouncementCloseReason = SnackbarCloseReason | 'closeButtonClick';
+export type AnnouncementCloseReason =
+	| SnackbarCloseReason
+	| 'closeButtonClick'
+	| 'linkClick';
 
 export interface AnnouncementProps {
 	open?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/Announcement/index.tsx
```diff
@@ -160,6 +171,11 @@ export const Announcement = React.forwardRef<HTMLDivElement, AnnouncementProps>(
 						<MUILinkWithTracking
 							href={linkHref}
 							eventName="Announcement: click on link"
+							onClick={(e) => {
+								if (onClose) {
+									onClose(e, 'linkClick');
+								}
```
myarmolinsky, line 177: nit
```suggestion
								onClose?.(e, 'linkClick');
```

src/theme.tsx
```diff
@@ -1763,6 +1763,17 @@ export const theme = createTheme({
 				},
 			},
 		},
+		MuiSnackbar: {
+			styleOverrides: {
+				anchorOriginBottomRight: ({ theme }) => ({
+					[theme.breakpoints.up('sm')]: {
+						// Make room for the 'need help' button
+						// TODO add a theme.spacing entry between 32 and 64
```
myarmolinsky, line 1771: Do we have more cases that would benefit from that? Aside from it being a breaking change (requiring incrementing all spacings above it by 1) I'm also hesitant to add things just for a special case

drskullster: not that I can think of right now @myarmolinsky but the leap from 32px to 64px is pretty big so I expect it will come in handy in the future

drskullster: here's the default MUI spacing grid: https://m2.material.io/design/layout/understanding-layout.html#material-measurements

each step is a 8px increment, it's not exponential in any way.

myarmolinsky: Yes of course! I was just commenting to raise the discussion

myarmolinsky (approved):

PR #169, opened 2025-01-09
Rjst

JSReds:


src/typings/common.d.ts
myarmolinsky, line 1: Is this something we can export? We use these in the UI too so it could be nice to not have to recreate them in the UI `common.d.ts`

src/components/RJST/oData/converter.spec.ts
myarmolinsky, line 1: What's up with this file? Why is it all commented out?

src/contexts/UiSharedComponentsContextProvider.tsx
```diff
@@ -0,0 +1,25 @@
+import { createContext } from 'react';
+import type { TFunction } from '../hooks/useTranslations';
+export interface UiSharedComponentsContextProviderInterface {
+	history: unknown;
```
myarmolinsky, line 4: We could fix this typing and name now to be `navigate: NavigateFunction`, right?

src/components/RJST/oData/converter.spec.ts
JSReds: my bad forgot to add tests, add them back

src/contexts/UiSharedComponentsContextProvider.tsx
JSReds: yep done in a separate commit

src/typings/common.d.ts
JSReds: I would keep the common.d.ts separated for now

myarmolinsky (approved):

PR #190, opened 2025-02-25
Secure boot

myarmolinsky:


src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -104,7 +107,13 @@ export const ApplicationInstructions = memo(function ApplicationInstructions({
 
 	const finalInstructions = [
 		'Use the form on the left to configure and download balenaOS for your new device.',
-		...interpolatedInstructions,
+		...(secureboot
+			? [
+					...interpolatedInstructions.slice(0, 6),
+					`${interpolatedInstructions[6].split('.').slice(0, 2)}.`,
```
thgreasi, line 113: Isn't there a better way to detect this? Like find the instruction step that includes x/y/z?

src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -190,7 +208,30 @@ const InstructionsItem = ({ node, index }: InstructionsItemProps) => {
 						return <p />;
 					},
 					p: ({ children }) => (
-						<p style={{ marginTop: 0, marginBottom: 0 }}>{children}</p>
+						<p style={{ marginTop: 0, marginBottom: 0 }}>
+							{children}
+							{secureboot && index === 7 && (
```
thgreasi, line 213: > Isn't there a better way to detect this? Like find the instruction step that includes x/y/z?
We should for sure also have a `TODO find a way to do this via contracts` or something.

src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -30,6 +30,9 @@ export const getUserOs = () => {
 	return 'Unknown';
 };
 
+const SECURE_BOOT_INSTRUCTION_START =
+	'Ensure there are no other USB keys are inserted. Power on the Generic x86_64 (GPT) with a keyboard connected.';
```
thgreasi, line 34: NIT: I would prefer storing the start of the part that needs to be replaced eg `Press the F10 key while BIOS`.
So it will be visible in the code that we are replacing the BIOS related section with UEFI.

src/components/DownloadImageDialog/ApplicationInstructions.tsx
```diff
@@ -190,29 +193,60 @@ const InstructionsItem = ({ node, index }: InstructionsItemProps) => {
...
+							instruction.startsWith(SECURE_BOOT_INSTRUCTION_START) ? (
+								<>
+									{SECURE_BOOT_INSTRUCTION_START}
+									<Box mt={2} borderLeft={1} borderColor="divider" pl={2}>
+										<Chip
+											sx={{ marginRight: 1 }}
+											label="Secure Boot"
+											color="purple"
+										/>
+										The device needs to be configured in secure boot setup mode.
```
thgreasi, line 208: Thinking out loud about a future where this comes from the contracts...
Would it work the same if we stored this as a markdown string like v, similarly to how we get the etcher link from contracts?
```
const UEFI_INSTRUCTIONS = `
The device needs to be configured in secure boot setup mode.
This depends on the UEFI/BIOS implementation, but in general, this involves resetting the UEFI settings to default configuration, configuring the device to boot in UEFI mode, setting the first boot option to the USB key and disabling the restoration of factory keys.
Save and Exit the UEFI/BIOS menu and the device should automatically reboot and begin the provisioning process.
[Read more on UEFI settings for secure boot and full disk encryption.](https://docs.balena.io/reference/OS/secure-boot-and-full-disk-encryption/generic-x86-64-gpt/#provision-the-device)`;
```

myarmolinsky: The markdown itself would work, but the chip being inline with the text would not. The chip would only be inline with the text if it is placed among plain text, not alongside a component such as `Box` or `Markdown`. The parent `Markdown` wrapper is not transforming this text automatically which is why we would need another wrapper, plus we style the extra instruction a bit differently with a left border making it stand out from the rest of the instruction that was not replaced

thgreasi (approved):
Remember to squash!

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -204,347 +251,550 @@ export const ImageForm = memo(function ImageForm({
...
-			<Button
-				onClick={() => {
-					setShowAdvancedSettings(!showAdvancedSettings);
+			<Dialog
+				open={showSecureBootConfirmationDialog}
+				onClose={() => {
+					setShowSecureBootConfirmationDialog(false);
+					dismissSecureBootWarning(false, dontShowSecureBootWarningAgain);
+					setDontShowSecureBootWarningAgain(false);
 				}}
```
JSReds, line 693: what about moving setShowSecureBootConfirmationDialog and setDontShowSecureBootWarningAgain inside dismissSecureBootWarning ? 
Why do we need to reset `dontShowSecureBootWarningAgain` to false ? if I reopen the modal I should see what I've selected right ?

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -204,347 +251,550 @@ export const ImageForm = memo(function ImageForm({
...
-							// TODO: find a better way to center the checkbox with the input only (without label)
-							height={54}
-						>
+						<FormControl>
+							<FormLabel id="secure-boot-and-full-disk-encryption-label">
+								<Stack direction="row" alignItems="center" gap={1}>
+									<Typography variant="titleSm">
+										Secure Boot and Full Disk Encryption
+									</Typography>
+									{/* TODO: Pick an actualy expiry date */}
```
JSReds, line 538: can we solve this TODO or we need some extra info ?

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -90,6 +104,9 @@ interface ImageFormProps {
 	) => void;
 }
 
+export const GENERIC_X86_SLUG = 'generic-amd64';
+export const GENERIC_X86_MINIMUM_SUPPORTED_SECUREBOOT_VERSION = '5.3.15';
```
JSReds, line 108: As far as I can see we can remove the export

src/components/DownloadImageDialog/ImageForm.tsx
myarmolinsky: We still don't know when we're shipping this. Once we pick a date to ship, I'll set an expiry date at least a month later

PR #191, opened 2025-02-27
DownloadImageDialog: Fix downloading the auto-selected release when it has variants

thgreasi:
An extra case that we forgot and used the
stripped version instead of the raw version.

Change-type: patch
See: https://balena.zulipchat.com/#narrow/channel/403752-channel.2Fsupport-help/topic/Orange.20pi.20Zero.20image.20download.20does.20not.20work/near/502224287

### DIFF
--- a/src/components/DownloadImageDialog/ImageForm.tsx
+++ b/src/components/DownloadImageDialog/ImageForm.tsx
@@ -136,11 +136,14 @@ export const ImageForm = memo(function ImageForm({
 	const handleVersionChange = useCallback(
 		(ver: typeof version) => {
 			if (!ver) {
-				const newVersion =
+				ver =
 					versionSelectionOpts.find((v) => v.isRecommended) ??
 					versionSelectionOpts[0];
-				onChange('version', newVersion?.value);
-				setVersion(newVersion);
+			}
+			if (ver == null) {
+				// This shouldn't really happen, but if it does, that's all we can do.
+				onChange('version', undefined);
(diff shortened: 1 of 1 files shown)


src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -136,11 +136,14 @@ export const ImageForm = memo(function ImageForm({
 	const handleVersionChange = useCallback(
 		(ver: typeof version) => {
 			if (!ver) {
-				const newVersion =
+				ver =
 					versionSelectionOpts.find((v) => v.isRecommended) ??
 					versionSelectionOpts[0];
-				onChange('version', newVersion?.value);
```
thgreasi, line 142: The early return here didn't allow the variant related code that we have later run :(

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -136,11 +136,14 @@ export const ImageForm = memo(function ImageForm({
...
-				setVersion(newVersion);
+			}
+			if (ver == null) {
+				// This shouldn't really happen, but if it does, that's all we can do.
+				onChange('version', undefined);
+				setVersion(undefined);
 				return;
```
thgreasi, line 147: The early return here didn't allow the variant related code that we have later run :(
So I just removed it.

myarmolinsky (approved):

PR #195, opened 2025-03-07
Use css variables in MUI

drskullster:
See: https://balena.fibery.io/Work/Project/Better-theme-integration-in-the-dashboard-1076

Change-type: minor


src/theme.tsx
```diff
@@ -521,8 +534,8 @@ export const theme = createTheme({
 					padding: 0,
 				},
 				action: ({ theme }) => ({
-					marginTop: `-${theme.spacing(2)}`,
-					marginBottom: `-${theme.spacing(2)}`,
+					marginTop: `calc(-1 * ${theme.spacing(2)})`,
+					marginBottom: `calc(-1 * ${theme.spacing(2)})`,
```
myarmolinsky, line 538: Why `calc` here and simply `-1` in `Announcement/index`?

PR #198, opened 2025-03-12
Allow users to reorder columns in settings menu

JSReds:
Change-type: minor

Screenshot 2025-03-26 at 16 03 33
Screenshot 2025-03-26 at 16 03 46
Screenshot 2025-03-26 at 16 04 15


src/components/RJST/components/Table/TableActions.tsx
```diff
@@ -28,6 +39,9 @@ export const TableActions = <T extends object>({
 	onColumnPreferencesChange,
 }: TableActionsProps<T>) => {
 	const [anchorEl, setAnchorEl] = React.useState<HTMLElement>();
+	const theme = useTheme();
+	const { state: analyticsState } = useAnalyticsContext();
+	const matches = useMediaQuery(theme.breakpoints.up('sm'));
```
myarmolinsky, line 44: nit:
```suggestion
	const mobile = useMediaQuery(theme.breakpoints.down('md'));
```

Like we do everywhere else?

src/components/RJST/components/Table/TableActions.tsx
```diff
@@ -41,25 +55,51 @@ export const TableActions = <T extends object>({
...
+			index: newIndex,
+		}));
+
+		onColumnPreferencesChange?.(updatedColumns, 'reorder');
+	};
+
 	return (
 		<>
-			<IconButton
+			<Button
```
myarmolinsky, line 90: nit: `ButtonWithTracking`?

src/components/RJST/components/Table/TableActions.tsx
```diff
@@ -68,35 +108,72 @@ export const TableActions = <T extends object>({
...
+										index={index}
+									>
+										{(item) => (
+											<MenuItem
+												ref={item.innerRef}
+												{...item.draggableProps}
+												sx={{ display: 'flex', alignItems: 'center', p: 0 }}
+											>
+												{analyticsState.featureFlags?.columnOrdering && (
+													<ListItemIcon
```
myarmolinsky, line 136: Why `ListItemIcon` instead of the icon alone? Necessary for `item.dragHandleProps`?

myarmolinsky (approved):

src/components/RJST/components/Table/TableActions.tsx
JSReds: mainly accessibility, in a list (MenuItem in this case) that's what is suggested to respect MUI's semantic structure

eg: https://mui.com/material-ui/react-menu/

src/components/RJST/components/Table/TableActions.tsx
JSReds: I can agree with doing always the same thing everywhere, but atm we use more theme.breakpoints.up('sm') then down('md') in ui-shared-compoennts. Let me open a different PR to add more consistency on that

src/components/RJST/components/Table/TableActions.tsx
JSReds: for this sepcific case we already send custom analytics, I would avoid sending them twice

https://github.com/balena-io-modules/ui-shared-components/blob/64d84d2a0bb48354956ca692f2709099222afb87/src/components/RJST/Lenses/types/table.tsx#L263

PR #204, opened 2025-03-17
VirtualizedAutocomplete: Fix console error caused by passing `loadNext` to `Autocomplete`

myarmolinsky:
Change-type: patch

### DIFF
--- a/src/components/VirtualizedAutocomplete/index.tsx
+++ b/src/components/VirtualizedAutocomplete/index.tsx
@@ -158,15 +158,17 @@ const VirtualizedAutocompleteBase = <
 	renderOption,
 	getOptionLabel,
 	value,
+	loadNext,
 	...props
 }:
-	| VirtualizedAutocompleteProps<
+	| (VirtualizedAutocompleteProps<
 			Value,
 			Multiple,
 			DisableClearable,
 			FreeSolo,
 			ChipComponent
-	  >
+			// This is to allow destructuring of loadNext from props
+	  > & { loadNext?: never })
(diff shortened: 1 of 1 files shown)


src/components/VirtualizedAutocomplete/index.tsx
```diff
@@ -231,7 +232,7 @@ const VirtualizedAutocompleteBase = <
 
 	return (
 		<Autocomplete<Value, Multiple, DisableClearable, FreeSolo, ChipComponent>
-			{...props}
+			{...omit(props, 'loadNext')}
```
thgreasi, line 235: How about just destrucuring it right above like `const { loadNext, ...passhroughProps } = props;`?
Lodash's Omit is one of the slow methods they offer, and given that it's just one prop, we can easily avoid it.

src/components/VirtualizedAutocomplete/index.tsx
```diff
@@ -158,15 +158,17 @@ const VirtualizedAutocompleteBase = <
...
-	| VirtualizedAutocompleteProps<
+	| (VirtualizedAutocompleteProps<
 			Value,
 			Multiple,
 			DisableClearable,
 			FreeSolo,
 			ChipComponent
-	  >
+			// This is to allow destructuring of loadNext from props
+	  > & { loadNext?: never })
```
thgreasi, line 171:
```suggestion
	  > & {
		// This is to allow destructuring of loadNext from props
		loadNext?: never;
	   })
```

thgreasi (approved):

PR #213, opened 2025-03-27
DownloadImageDialog: Move development variant warning to a callout

myarmolinsky:
Change-type: patch

### DIFF
--- a/src/components/DownloadImageDialog/ImageForm.tsx
+++ b/src/components/DownloadImageDialog/ImageForm.tsx
@@ -8,7 +8,6 @@ import {
 	FormControlLabel,
 	FormLabel,
 	InputAdornment,
-	InputLabel,
 	Radio,
 	RadioGroup,
 	TextField,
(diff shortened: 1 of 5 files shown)


src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -466,84 +468,96 @@ export const ImageForm = memo(function ImageForm({
...
+								onChange={(event) => {
+									onChange('appUpdatePollInterval', event.target.value);
 								}}
-							>
-								<ArticleIcon sx={{ ml: 1, fontSize: '1.15rem' }} />
-							</MUILinkWithTracking>
-						</FormLabel>
+							/>
+						</FormControl>
+						<InputLabel htmlFor="provision-key-name" sx={{ my: 2 }}>
```
drskullster, line 524: I think we can get rid of `my: 2` (I don't see a reason for it to be there). Also we could probably use the `label` property of the `TextField` component instead (true for other components here as well I think)

drskullster (approved):

PR #214, opened 2025-03-28
Replace deprecated `react-beautiful-dnd` with `dnd-kit`

JSReds:
Change-type: patch


src/components/RJST/components/Table/TableActions.tsx
```diff
@@ -33,6 +29,45 @@ interface TableActionsProps<T> {
...
+			<FormControlLabel
+				sx={{ flex: 1, width: '100%', m: 0 }}
+				control={
+					<Checkbox
+						sx={{ m: 0 }}
+						onClick={() => handleColumnSelection(column)}
+						checked={column.selected}
+					/>
+				}
+				label={typeof column.label === 'string' ? column.label : column.title}
```
myarmolinsky, line 65:
```suggestion
				label={column.label ?? column.title}
```

src/components/RJST/components/Table/TableActions.tsx
```diff
@@ -43,46 +78,61 @@ export const TableActions = <T extends object>({
...
+
+		if (oldIndex === -1 || newIndex === -1) {
+			return;
+		}
+
+		const reordered = arrayMove(columns, oldIndex, newIndex).map((c, i) => ({
+			...c,
+			index: i,
 		}));
+		console.log(active, over, reordered);
```
myarmolinsky, line 133: console.log

src/components/RJST/components/Table/TableActions.tsx
JSReds: this is not the same thing, we are checking if the actions is a JSX element or a string as in this dropdown we only want to show the string and not the entire JSX element that might contain a tootlip or other stuff

myarmolinsky (approved):

PR #220, opened 2025-04-10
Add Horizontal Bar Chart component

drskullster:
Replaces our current chart without the external dependency

See https://balena.fibery.io/Work/Project/Replace-the-external-library-for-the-horizontal-bar-chart-with-a-SVG-based-solution-1316

Change-type: minor

### DIFF
--- a/src/components/HorizontalBarChart/index.tsx
+++ b/src/components/HorizontalBarChart/index.tsx
@@ -1,0 +1,103 @@
+import { Stack, styled, Tooltip, Typography } from '@mui/material';
+import { token } from '../../utils/token';
+import { useTranslation } from '../../hooks/useTranslations';
+import { useMemo } from 'react';
+
+export interface HorizontalBarChartProps {
+	items: Array<{ color: string; title: string; count: number }>;
+	resourceLabel?: string;
+	resourceLabelPlural?: string;
+}
+
+const Bar = styled('div')({
+	width: '100%',
+	height: 24,
+	display: 'flex',
+	borderRadius: token('shape.borderRadius.sm'),
+	overflow: 'hidden',
+	gap: 2,
+});
+
+const BarItem = styled('span')({
+	minWidth: 3,
+});
+
+const Legend = styled('ul')({
+	display: 'inline',
+	listStyle: 'none',
+	padding: 0,
+	margin: 0,
+});
+
+const LegendItem = styled('li')({
+	display: 'inline-flex',
+	marginRight: token('spacing.2'),
+	alignItems: 'center',
+	'&:before': {
+		content: '""',
+		backgroundColor: 'currentColor',
+		borderRadius: '50%',
+		width: '10px',
+		height: '10px',
+		display: 'inline-block',
+		marginRight: token('spacing.1'),
+	},
+});
+
+export const HorizontalBarChart = ({
+	items,
+	resourceLabel,
+	resourceLabelPlural,
+}: HorizontalBarChartProps) => {
+	const { t } = useTranslation();
+	const totalItems = useMemo(
+		() => items.reduce((sum, item) => sum + item.count, 0),
+		[items],
+	);
+	const filteredItems = useMemo(
+		() => items.filter((item) => item.count > 0),
+		[items],
+	);
+	const getTotalResourcesLabel = (count: number) => {
+		if (resourceLabel && resourceLabelPlural) {
+			return t('labels.total_resources', {
+				count,
+				resource: count > 1 ? resourceLabelPlural : resourceLabel,
+			});
+		} else {
+			return count;
+		}
+	};
+
+	return (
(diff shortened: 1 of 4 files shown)


src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const HorizontalBarChart = ({ items }: HorizontalBarChartProps) => {
+	const { t } = useTranslation();
+	const totalItems = useMemo(
+		() => items.reduce((sum, item) => sum + item.count, 0),
+		[items],
+	);
+	const getTotalDevicesLabel = (count: number) =>
+		count > 1
+			? t('labels.total_devices_other', { count })
+			: t('labels.total_devices_one', { count });
```
myarmolinsky, line 54: I don't like that we're hardcoding the resource (`device`) for the label. We should add a prop to the `HorizontalBarChart` component for specifying the resource (i.e. `resource: string, resourcePlural: string`). If you want to make the prop optional, we should just display the count when the `resource` is undefined.

src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+						<Tooltip
+							key={item.title}
+							title={`${item.title}: ${getTotalDevicesLabel(item.count)}`}
+						>
+							<BarItem
+								sx={{
+									backgroundColor: item.color,
+									width: `${(item.count / totalItems) * 100}%`,
+								}}
+							></BarItem>
```
myarmolinsky, line 71: ? or am I missing something?
```suggestion
						>
							<BarItem
								sx={{
									backgroundColor: item.color,
									width: `${(item.count / totalItems) * 100}%`,
								}}
							/>
```

src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+	width: '100%',
+	height: 24,
+	display: 'flex',
+	borderRadius: token('shape.borderRadius.sm'),
+	overflow: 'hidden',
+	gap: 2,
+});
+
+const BarItem = styled('span')({
+	minWidth: 3,
```
myarmolinsky, line 20: What about when width is 0%?

myarmolinsky: > .filter((item) => item.count > 0)

Nevermind I see we filter those out

src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+		height: '10px',
+		display: 'inline-block',
+		marginRight: token('spacing.1'),
+	},
+});
+
+export const HorizontalBarChart = ({ items }: HorizontalBarChartProps) => {
+	const { t } = useTranslation();
+	const totalItems = useMemo(
+		() => items.reduce((sum, item) => sum + item.count, 0),
```
myarmolinsky, line 48: IIRC either Thodoris, Page, or both hate `reduce` because it is less performant than a for loop 😅

src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+									width: `${(item.count / totalItems) * 100}%`,
+								}}
+							></BarItem>
+						</Tooltip>
+					))}
+			</Bar>
+
+			<Legend>
+				{items
+					.filter((item) => item.count > 0)
```
myarmolinsky, line 78: So we filter out items with count <= 0 twice. Let's move that to a useMemo

src/components/HorizontalBarChart/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+const Legend = styled('ul')({
+	display: 'inline',
+	listStyle: 'none',
+	padding: 0,
+	margin: 0,
+});
+
+const LegendItem = styled('li')({
+	display: 'inline-flex',
+	marginRight: token('spacing.2'),
```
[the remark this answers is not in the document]
drskullster: @myarmolinsky the shortcuts for spacing, colors etc don't work in `styled` functions unfortunately, only in `sx` and component props. 

I could inject the `theme` object if you prefer, but `token()` writes more naturally to me already. Whatever you prefer!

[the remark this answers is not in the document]
drskullster: > the shortcuts for spacing, colors etc don't work in styled functions unfortunately, only in sx and component props.

which may be an argument to avoid them honestly, since we already started replacing them for colors.

src/components/HorizontalBarChart/index.tsx
drskullster: Not to call anyone out but I found this in the UI 😁
image

myarmolinsky (approved):

PR #227, opened 2025-04-23
DownloadImageDialog: Fix the variant selection for non-unified OS releases

thgreasi:
Change-type: patch
See: https://balena.fibery.io/Work/Project/DownloadImageDialog-Fix-the-variant-selection-for-non-unified-OS-releases-1405

### DIFF
--- a/src/components/DownloadImageDialog/ImageForm.tsx
+++ b/src/components/DownloadImageDialog/ImageForm.tsx
@@ -86,10 +86,11 @@ interface ImageFormProps {
 	model: DownloadImageFormModel;
 	hasEsrVersions?: boolean;
 	onSelectedOsTypeChange: (osType: string) => void;
-	onChange: (
-		key: keyof DownloadImageFormModel,
-		value: DownloadImageFormModel[keyof DownloadImageFormModel],
-	) => void;
+	// onChange needs to be passing a complete object of changes
+	// since when having multipele onChange call, the subsequent ones
+	// can have issues b/c of still using the original state, before the
+	// previous onChange calls.
+	onChange: (obj: Partial<DownloadImageFormModel>) => void;
 }
 
 export const ImageForm = memo(function ImageForm({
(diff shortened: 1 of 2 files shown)


src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -395,7 +404,7 @@ export const ImageForm = memo(function ImageForm({
 						value={model.network}
 						name="network"
 						onChange={(event) => {
-							onChange('network', event.target.value);
+							onChange({ network: event.target.value as 'ethernet' | 'wifi' });
```
myarmolinsky, line 407:
```suggestion
							onChange({ network: event.target.value as DownloadImageFormModel['network'] });
```

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -124,29 +121,41 @@ export const ImageForm = memo(function ImageForm({
...
-				} else {
+				if (!rawVersionForVariant) {
 					handleVariantChange(variant === 'dev' ? 'prod' : 'dev');
+					return;
 				}
-				return;
+				newState.version = rawVersionForVariant;
+			} else {
+				newState.version = ver?.rawVersion;
 			}
```
myarmolinsky, line 157: Move the `newState.version = ver?.rawVersion;` above the `if` and get rid of the `else`

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -124,29 +121,41 @@ export const ImageForm = memo(function ImageForm({
...
+			setVariant(newVariant);
+			const newState: Partial<DownloadImageFormModel> = {
+				developmentMode: newVariant === 'dev',
+			};
+			if (version?.hasPrebuiltVariants) {
+				const rawVersionForVariant = version.rawVersions[newVariant];
+				if (rawVersionForVariant != null) {
+					newState.version = rawVersionForVariant;
+				}
+			}
```
myarmolinsky, line 134: Please double check that we do not have a useEffect or something somewhere that is triggered by `developmentMode` changing that already does this so that we do not duplicate this

thgreasi: We did't have any, and that's why this issue is happening at the first place :)

src/components/DownloadImageDialog/ImageForm.tsx
thgreasi: We can't do that b/c `rawVersion` only exists in the union type when `hasPrebuiltVariants: false`, so before the `if` TS doesn't allow us to access it.

src/components/DownloadImageDialog/ImageForm.tsx
```diff
@@ -124,11 +124,20 @@ export const ImageForm = memo(function ImageForm({
...
+		(newVariant: typeof variant) => {
+			setVariant(newVariant);
+			onChange('developmentMode', newVariant === 'dev');
+			if (version?.hasPrebuiltVariants) {
+				const rawVersionForVariant = version.rawVersions[newVariant];
+				if (
+					rawVersionForVariant != null &&
+					rawVersionForVariant !== version.value
+				) {
+					onChange('version', rawVersionForVariant);
```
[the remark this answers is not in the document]
myarmolinsky: Better to put a comment explaining why this is necessary for non-unified OS releases than just a comment saying to test them

PR #231, opened 2025-04-30
Remove role inheritance

myarmolinsky:
See: https://balena.fibery.io/Work/Project/Change-Membership-Role-Permissions-to-not-have-inheritance-in-the-UI-1393

### DIFF
--- a/src/components/RJST/schemaOps.ts
+++ b/src/components/RJST/schemaOps.ts
@@ -11,7 +11,7 @@ import type { PineFilterObject } from './oData/jsonToOData';
 
 export interface RJSTBaseResource<T> {
 	id: number;
-	__permissions: Permissions<T>;
+	__permissions: Array<Permissions<T>>;
 }
 
 type XHeaderLink = {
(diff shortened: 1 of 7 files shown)


src/components/RJST/schemaOps.ts
```diff
@@ -11,7 +11,7 @@ import type { PineFilterObject } from './oData/jsonToOData';
 
 export interface RJSTBaseResource<T> {
 	id: number;
-	__permissions: Permissions<T>;
+	__permissions: Array<Permissions<T>>;
```
thgreasi, line 14: How about returning the combined permissions? (ie keep it as `__permissions: Permissions<T>;`)
I think that would be semantically more appropriate.
ie: autoUI will not care about your permission groups (that the API knows), it will only care about what your final permission set is at the end of the day.

In an ideal world I expect we would have a helper for merging permissions which we would be exporting so that the UI can use it inside its `__permissions (_entity: T, context: RJSTContext<T>): Permissions<T>` callbacks.

If we do it that way, then I expect that we will not need any changes in RJST (since it's already agnostic enough) other than perhaps defining and exporting the new `mergePermissons` (or something) helper function.
The UI will be then calling `mergePermissons` in a memoized way so that everything stays performant and without the need of extra loops.

myarmolinsky:
Closing in favor of creating and using `mergePermissons` in the UI

PR #242, opened 2025-07-21
Material 7

myarmolinsky:


package.json
```diff
@@ -17,15 +17,15 @@
     "@emotion/styled": "^11.10.6",
     "@fortawesome/free-solid-svg-icons": "^6.5.1",
     "@fortawesome/react-fontawesome": "^0.2.0",
-    "@mui/lab": "6.0.0-beta.24",
-    "@mui/material": "^6.3.1",
-    "@mui/styled-engine-sc": "^6.3.1",
-    "@mui/x-data-grid": "^7.23.5",
+    "@mui/lab": "7.0.0-beta.14",
```
JSReds, line 20: is it still necessary ? I mean do we still need to keep the beta ?

myarmolinsky: That PR you linked is not the reason for the beta, it is the reason for the version not having a `^`. And I believe we have agreed that it's better practice in general not to have a `^` for beta versions.

The lab package itself is all betas, it only changes major versions aside from that whenever MUI base is bumped. We use the lab package for the Tabs components. As long as those have not graduated to MUI base (or been removed/canceled), we cannot remove the lab package

src/components/Form/FieldTemplates/ObjectFieldTemplate.tsx
```diff
@@ -1,4 +1,4 @@
-import Grid from '@mui/material/Grid';
+import { Grid } from '@mui/material';
```
JSReds: just in case, this component was created checking theirs and adding some sx rules. To keep consistency between Object templates fields I would keep this as default but not strong opinion on that.

here their component https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/mui/src/ObjectFieldTemplate/ObjectFieldTemplate.tsx

JSReds (approved):

PR #294, opened 2026-06-10
Add Collapsible component and add some properties to the DropDownButton component

myarmolinsky:
Change-type: patch
Connects-to: https://github.com/balena-io/balena-ui/pull/7419
Fibery: https://balena.fibery.io/Work/Project/Helios-MLP-Release-Plan-1704
Fibery: https://balena.fibery.io/Work/Improvement/Add-supervisor-services-to-UI-4331
Fibery: https://balena.fibery.io/Work/Project/Show-Supervisor-services-on-the-dashboard-2642

### DIFF
--- a/src/components/Collapsible/index.tsx
+++ b/src/components/Collapsible/index.tsx
@@ -1,0 +1,130 @@
+import { faChevronUp } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import {
+	Accordion,
+	accordionSummaryClasses,
+	AccordionSummary,
+	AccordionDetails,
+} from '@mui/material';
+import type {
+	AccordionDetailsProps,
+	AccordionProps,
+	AccordionSummaryProps,
+} from '@mui/material';
+import { useState } from 'react';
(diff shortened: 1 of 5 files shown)


src/components/Collapsible/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+	detailsProps,
+	summaryProps,
+	collapsedLocalStorageKey,
+	...props
+}: CollapsibleProps) => {
+	const [collapsed, setCollapsed] = useState(
+		collapsedLocalStorageKey
+			? (getFromLocalStorage(collapsedLocalStorageKey) ?? initiallyCollapsed)
+			: initiallyCollapsed,
+	);
```
JSReds: Shouldn't we consider the expanded prop as well? If not, I'd omit it from the type since the component is managing the expanded state internally.

NIT: I'd use `isCollapsed` as name for the boolean state, since that's the convention we use elsewhere too.

https://mui.com/material-ui/api/accordion/#:~:text=height%20when%20expanded.-,expanded,-bool

myarmolinsky: Fair point. I've omitted it for now and left a comment that we should revisit it if the use case arises where the parent needs to control the expanded state

JSReds (approved):

PR #295, opened 2026-06-16
Switch from custom `Dictionary` type to built in `Record` typing

Page-:
This reduces the number of custom typings we maintain/need to know about

Change-type: patch


src/typings/common.d.ts
```diff
@@ -1,9 +1,6 @@
 type MaybePromise<T> = T | Promise<T>;
 
 type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;
-interface Dictionary<T> {
```
Page-, line 4: This module was previously relying on this always being globally available which meant the generated definitions included `Dictionary<...>` verbatim without actually defining `Dictionary` at any point, which would result in the whole type being created as `any` unless the consumer also declares a global `Dictionary` type (and you'd better hope the definition matches..)

myarmolinsky (approved):

PR #303, opened 2026-07-03
Fix password widget issues when pasting or using a password manager

drskullster:
Change-type: patch


src/components/Form/Widgets/PasswordWidget.tsx
```diff
@@ -35,7 +35,7 @@ export const PasswordWidget = ({
 	const [showPassword, setShowPassword] = useState(false);
 
 	const change = ({ target: { value: val } }: any) => {
-		onChange(value === '' ? options.emptyValue : val);
+		onChange(val);
 	};
```
JSReds, line 39: I guess we can just remove this function then and pass the onChange directly in the component

drskullster: I thought of that as well @JSReds but we're destructuring `event.target.value` so we would need to change the `onChange` prop of every consumer of the widget right?

JSReds: I mean, we could just inline it instead of defining a separate function, as we do for the other handlers (`onBlur`, `onFocus`, etc.).

```
onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
	change(event.target.value);
}}
```

JSReds (approved):

